PR #1, opened 2020-03-24
Create BMP.md

brmcdoug:
I've gotten a start on a BMP.md doc.  Does this format work and give you what you're looking for?

### DIFF
--- a/BMP.md
+++ b/BMP.md
@@ -27,83 +27,83 @@ https://www.snas.io/docs/
 
 #### BMP peer message:
 ```
-1: (action): up/down
+1: (action): string (up/down)
 5: (name): 
 6: (remote_bgp_id): 
 7: (router_ip): 
 8: (timestamp): 
-9: (remote_asn): 
+9: (remote_asn): int
 10: (remote_ip): 
 11: (peer_rd): 
-13: (local_asn): 
+13: (local_asn): int
 14: (local_ip): 
 16: (local_bgp_id):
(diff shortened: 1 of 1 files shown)


BMP.md
```diff
@@ -0,0 +1,64 @@
+# BGP Monitoring Protocol (BMP)
+
+https://tools.ietf.org/html/rfc7854
+
+## BMP message types and GoBMP:
```
sbezverk: When displayed as reach text, it is bit hard to read as everything is displayed on the same line. Could you please check the formatting?

BMP.md
```diff
@@ -0,0 +1,64 @@
...
+GoBMP will not need to collect and parse the following BMP message types:
```
sbezverk, line 6: In both statements you say GoBMP will **not**, I suspect one of these should be without "not" right?

BMP.md
```diff
@@ -0,0 +1,245 @@
...
+l3vpn
+evpn
+```
+With much gratitude we are using OpenBMP's message parsing as a template:
+
+https://www.snas.io/docs/
+
+#### BMP peer message:
+```
+1: (action): up/down
```
sbezverk: Please confirm if `action` field is a string with actual strings "up" or "down" or it is numerical enumeration where for example 0 means up and 1 means down.

BMP.md
brmcdoug: fixed, thanks

PR #13, opened 2020-04-06
Topology - PeerChange

brmcdoug:


pkg/topology/arangodb/arangodb-peer-change.go
```diff
@@ -0,0 +1,20 @@
...
+	bgp_peer_document := &arangoDB.BGPPeer{
+		RouterID:	message.LocalBGPID
+		LocalIP:	message.LocalIP
+		LocalASN:	message.LocalASN
+		PeerIP:		message.RemoteIP
+		PeerBGPID:	message.RemoteBGPID
+		PeerASN:	message.RemoteASN
+	}
+	if err := m.arangoDB.Upsert(bgp_peer_document); err != nil {
+		fmt.Println("Encountered an error while upserting the bgp peer document", err)
```
sbezverk: let's use glog.Errorf() for error reporting, you can find some examples in the repo.

pkg/topology/arangodb/arangodb-peer-change.go
```diff
@@ -0,0 +1,20 @@
+package arangodb
+
+func (m *arangoDB) peerChangeHandler(obj *message.PeerStateChange) {
```
sbezverk: other methods use (a *arangoDB) instead of (m *arangoDB) it is a good practice to use the same, otherwise go linter will complain.

pkg/topology/arangodb/arangodb-peer-change.go
```diff
@@ -0,0 +1,20 @@
...
+	// creates a new BGP peer document in Arango's BGPPeer vertex collection
+	bgp_peer_document := &arangoDB.BGPPeer{
+		RouterID:	message.LocalBGPID
+		LocalIP:	message.LocalIP
+		LocalASN:	message.LocalASN
+		PeerIP:		message.RemoteIP
+		PeerBGPID:	message.RemoteBGPID
+		PeerASN:	message.RemoteASN
+	}
+	if err := m.arangoDB.Upsert(bgp_peer_document); err != nil {
```
sbezverk: If Upsert is methof provided by arango go client, then it is not yet available as I still need to add arango client instantiation.  but if this method is/will be developed by you, then you would call it **a.Upsert**

pkg/topology/arangodb/arangodb-peer-change.go
```diff
@@ -0,0 +1,20 @@
...
+	// creates a new BGP peer document in Arango's BGPPeer vertex collection
+	bgp_peer_document := &arangoDB.BGPPeer{
+		RouterID:	message.LocalBGPID
+		LocalIP:	message.LocalIP
+		LocalASN:	message.LocalASN
+		PeerIP:		message.RemoteIP
+		PeerBGPID:	message.RemoteBGPID
+		PeerASN:	message.RemoteASN
+	}
+	if err := a.Upsert(bgp_peer_document); err != nil {
```
sbezverk, line 15: I do not see a direct call for Upsert. Did you try to compile your code? I suspect you had a helper function in your old topology which was used to do Upsert..  Please send me a line in the old topology where you used it. I will try to figure out.
Another thing is you need to check for the collection, open collection and then store it in it.

pkg/topology/arangodb/arangodb-peer-change.go
```diff
@@ -0,0 +1,20 @@
+package arangodb
+
+func (a *arangoDB) peerChangeHandler(obj *message.PeerStateChange) {
+	glog.Infof("><SB> peer change handler - update db with peer change message data")
+
+	// creates a new BGP peer document in Arango's BGPPeer vertex collection
+	bgp_peer_document := &arangoDB.BGPPeer{
```
sbezverk, line 7: Please use camel case for variables names example: `bgpPeerDocument`

pkg/topology/arangodb/arangodb-peer-change.go
brmcdoug: ok, i think it is in one of these:
https://github.com/cisco-ie/jalapeno/blob/master/processors/topology/handler/arango_handler.go
https://github.com/cisco-ie/jalapeno/blob/master/processors/topology/database/arango.go
https://github.com/cisco-ie/jalapeno/blob/master/processors/topology/database/database.go
https://github.com/cisco-ie/jalapeno/blob/master/processors/topology/database/epe_node.go

pkg/topology/arangodb/arangodb-peer-change.go
brmcdoug: done

sbezverk:
@brmcdoug  Please use this sample:
https://github.com/sbezverk/gobmp/blob/master/pkg/topology/arangodb/sample-peer.go
Also please make sure that code pushed gets compiled without errors.

sbezverk:
closing this pull request as topology related development is done in a different repo.

PR #35, opened 2020-04-20
recommit

SergeKrier:


pkg/bgp/extended-community.go
```diff
@@ -48,14 +48,23 @@ func (ext *ExtCommunity) String() string {
 		prefix = "sas="
 	case 0x0a:
 		prefix = "l2i="
+	case 0x0b:
+		prefix = "color="
```
sbezverk, line 52: Just a question. Is there any convention/abbreviation which should be used for identification of type of communities?

pkg/bgp/extended-community.go
```diff
@@ -48,14 +48,23 @@ func (ext *ExtCommunity) String() string {
...
 	}
+
+
```
sbezverk: Please remove blank lines.

pkg/bgp/extended-community.go
SergeKrier: no official ones that I know, but would be good practice to use those displayed in XR as ref

pkg/bgp/extended-community.go
```diff
@@ -109,6 +118,12 @@ func makeExtCommunity(b []byte) (*ExtCommunity, error) {
 		ext.SubType = &st
 		l = 6
 		p++
+	case 3: 
+		st := uint8(b[p])
+                ext.SubType = &st
+		l = 6
+		p+=3
```
sbezverk: is only 4 last bytes used for color community?

SergeKrier: correct, first 2 bytes are reserved, although following draft will use 2 bits (co-flags) : ietf-spring-segment-routing-policy ... to be added later

PR #92, opened 2020-08-21
ext community refactor

sbezverk:
Signed-off-by: Serguei Bezverkhi <sbezverk@cisco.com>

This PR refactors Extended Community processing to make it more modular and extensible

### DIFF
--- a/pkg/bgp/extended-community-const.go
+++ b/pkg/bgp/extended-community-const.go
@@ -1,0 +1,109 @@
+package bgp
+
+const (
+
+	// ECPRouteTarget extended community prefix for for Route Target	[RFC4360]
+	ECPRouteTarget = "rt="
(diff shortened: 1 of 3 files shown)


SergeKrier:
what's the point of using 3 letters acronym (sometimes cryptic), for the extended communities, can we use more human readable format ?

sbezverk:
@SergeKrier I was trying to follow as much as I can existing conventions. I am ok to revise naming, we just need to make sure it will not break anything which is already using certain prefixes. @brmcdoug  Could you please check proposed changes and let me know if it will break anything on your side in topology code.

brmcdoug:
I don't think the changes will break anything on the topology side. Most of those items are not yet in use, or like in the case of l3vpn, we already have "rt="

SergeKrier (approved):
I would have used ExCom prefix instead of EC but would be probably too long, ok like this

PR #126, opened 2020-11-11
Srte fixes

mirlos-pt:
The first commit is a fix to the 12-bit mt_id, wasn't noticed yesterday.
The isprepolicy name doesn't match the other is_flags.

An LSNode entry is always in one protocol, so let's make the area_id attribute consistent with LSLink and LSPrefix. While I didn't see either area id being set in LSSRv6SID, I kept the unified attribute anyway.
Some doc fixes.
The last one is an auto-generated change.

### DIFF
--- a/pkg/base/multitopologyidentifier.go
+++ b/pkg/base/multitopologyidentifier.go
@@ -25,7 +25,7 @@ func UnmarshalMultiTopologyIdentifierTLV(b []byte) ([]*MultiTopologyIdentifier, 
 	for i := 0; i < int(len(b)/2); i++ {
 		m := &MultiTopologyIdentifier{}
 		d := binary.BigEndian.Uint16(b[p : p+2])
-		m.MTID = d & 0x0ffff
+		m.MTID = d & 0x0fff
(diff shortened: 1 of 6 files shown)


sbezverk:
@mirlos-pt so the idea of using a common (between ISIS and OSPF) `AreaID`,  is that higher level application will parse Area ID appropriately based on a protocol id.
@brmcdoug Do you see any side effects especially in apps which might already be using ISIS and OSPF specific Area ID?

pkg/base/multitopologyidentifier.go
```diff
@@ -25,7 +25,7 @@ func UnmarshalMultiTopologyIdentifierTLV(b []byte) ([]*MultiTopologyIdentifier,
 	for i := 0; i < int(len(b)/2); i++ {
 		m := &MultiTopologyIdentifier{}
 		d := binary.BigEndian.Uint16(b[p : p+2])
-		m.MTID = d & 0x0ffff
+		m.MTID = d & 0x0fff
```
sbezverk, line 28: Good catch, thank you!

PR #129, opened 2020-11-27
Add protocol_id to LSSRv6SID

mirlos-pt:
A minor omission in the struct, where others include the numeric protocol ID along with the string.


sbezverk:
@mirlos-pt Thank you for catching it.

PR #134, opened 2020-12-09
Resolve Flex-Algo Definition SubTLVs

mirlos-pt:
I think this is a better representation of the optional FAD attributes. Thanks.


sbezverk (approved):

sbezverk:
@mirlos-pt looks good to me. I hope nobody other than you was using these attributes yet, so we are not breaking things for others
@brmcdoug we need to update topology to reflect change in the message structure proposed by @mirlos-pt

Issue #159, opened 2021-03-30
Restructure LSPrefix message Prefix Attribute TLVs


sbezverk:
To make Prefix Attribute TLVs to follow a structure outlined in RFC

PR #163, opened 2021-03-30
Prefix attr tlvs and flags

sbezverk:
closes: #159

### DIFF
--- a/pkg/bgpls/bgpls-srv6-policy-state.go
+++ b/pkg/bgpls/bgpls-srv6-policy-state.go
@@ -352,7 +352,7 @@ func (a *SRAffinityConstraint) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, t); err != nil {
 		return err
 	}
-	a = t
+	*a = *t
 
 	return nil
 }
@@ -402,7 +402,7 @@ func (s *SRSRLGConstraint) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, t); err != nil {
 		return err
 	}
-	s = t
+	*s = *t
 
 	return nil
 }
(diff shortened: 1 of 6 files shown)


pkg/bgpls/bgpls-srv6-policy-state.go
```diff
@@ -352,7 +352,7 @@ func (a *SRAffinityConstraint) UnmarshalJSON(b []byte) error {
 	if err := json.Unmarshal(b, t); err != nil {
 		return err
 	}
-	a = t
+	*a = *t
```
mirlos-pt, line 355: These changes do not actually belong to the commit, but I've done worse.

pkg/bgpls/prefix-attr-tlv.go
```diff
@@ -0,0 +1,323 @@
...
+			LSPrefixSID    []*sr.PrefixSIDTLV `json:"ls_prefix_sid,omitempty"`
+			SourceRouterID string             `json:"source_router_id,omitempty"`
+		}{
+			LSPrefixSID:    p.LSPrefixSID,
+			SourceRouterID: p.SourceRouterID,
+		})
+	}
+}
+
+func (p *PrefixAttrTLVs) UnmarshalJSON(b []byte) error {
```
mirlos-pt, line 79: Is it possible to pass the ProtoID, and then instead of guessing by the keys present, mapping to the correct PrefixAttrFlags struct? That way, you don't require the identifying flag keys to be present if they're false.

mirlos-pt (approved):

pkg/bgpls/bgpls-srv6-policy-state.go
sbezverk: Totally agree, bad practice, just worried to forget it.

pkg/bgpls/prefix-attr-tlv.go
sbezverk: `UnmarshalJSON(b []byte) error` unfortunately this is the call pattern must be used so unmarshalling of the data structure can be called by other packages.

PR #181, opened 2021-12-17
fix UnmarshalBGPOpenMessage

fujita:
The minimum length of the OPEN message is 29 octets (including the
message header). The callers of this function exclude the marker
bytes.

Signed-off-by: FUJITA Tomonori <fujita.tomonori@gmail.com>


sbezverk:
@fujita  Thank you for your PR

Issue #186, opened 2022-01-11
Hit segmentation fault when new bmp connection is made


lordofire:
Found a segmentation fault while gobmp is running, and a switch is connected to gobmp.
```
I0111 21:56:07.285544       1 gobmpsrv.go:33] Starting gobmp server on [::]:5000, intercept mode: false
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x697d8e]

goroutine 18 [running]:
github.com/sbezverk/gobmp/pkg/message.(*producer).marshalAndPublish(0xc000020d00, {0x7705e0, 0xc0000de300}, 0x1, {0xc0000b46e0, 0x20, 0x20}, 0x0)
	/home/sbezverk/projects/go/workspace/gobmp/pkg/message/route-monitor.go:93 +0x12e
github.com/sbezverk/gobmp/pkg/message.(*producer).producePeerMessage(0xc000020d00, 0x0, {0xc0000da700, {0x770120, 0xc00005c9b0}})
	/home/sbezverk/projects/go/workspace/gobmp/pkg/message/peer.go:100 +0xc9a
github.com/sbezverk/gobmp/pkg/message.(*producer).producingWorker(0xc0000f20c0, {0xc0000da700, {0x770120, 0xc00005c9b0}})
	/home/sbezverk/projects/go/workspace/gobmp/pkg/message/producer.go:44 +0x55
created by github.com/sbezverk/gobmp/pkg/message.(*producer).Producer
	/home/sbezverk/projects/go/workspace/gobmp/pkg/message/producer.go:33 +0x4c
```
After gdb into the system, I found the following information:
```
(gdb) bt
#0  0x0000000000697d8e in github.com/sbezverk/gobmp/pkg/message.(*producer).marshalAndPublish (p=0xc00001ed80, msg=..., msgType=10, 
    hash=..., debug=false) at /home/sbezverk/projects/go/workspace/gobmp/pkg/message/route-monitor.go:93
#1  0x000000000069533a in github.com/sbezverk/gobmp/pkg/message.(*producer).producePeerMessage (p=0xc00001ed80, op=0, msg=...)
    at /home/sbezverk/projects/go/workspace/gobmp/pkg/message/peer.go:100
#2  0x00000000006973d5 in github.com/sbezverk/gobmp/pkg/message.(*producer).producingWorker (p=0xc000112700, msg=...)
[...]

PR #187, opened 2022-01-12
add error check for all types of publishers

sbezverk:
Signed-off-by: Serguei Bezverkhi <sbezverk@cisco.com>
closes: #186

### DIFF
--- a/cmd/gobmp/gobmp.go
+++ b/cmd/gobmp/gobmp.go
@@ -78,24 +78,21 @@ func main() {
 	case "file":
 		publisher, err = filer.NewFiler(file)
 		if err != nil {
-			glog.Errorf("fail to initialize file publisher with error: %+v", err)
-			glog.Errorf("restarting gobmp...")
+			glog.Errorf("failed to initialize file publisher with error: %+v", err)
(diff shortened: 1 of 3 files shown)


cmd/gobmp/gobmp.go
```diff
@@ -76,9 +76,21 @@ func main() {
 	var err error
 	switch strings.ToLower(dump) {
 	case "file":
-		publisher = filer.NewFiler(file)
+		publisher, err = filer.NewFiler(file)
+		if err != nil {
+			glog.Errorf("fail to initialize file publisher with error: %+v", err)
+			glog.Errorf("restarting gobmp...")
```
lordofire: Is this restarting or just killing?

lordofire (approved):
lgtm with nits, thanks!

cmd/gobmp/gobmp.go
sbezverk: well, since in most of cases it runs as a deployment in kubernetes then when pod terminates, it gets restarted automatically.  I am removing second line just to avoid confusion.

PR #200, opened 2022-06-11
gobmp.go cleanup

taktv6:
- Cleaning up signal handling
- Dropping some obvious comments
- Using net.JoinHostPort instead of fmt.Sprintf
- Removing unnecessary os.Exit() calls


sbezverk (changes requested):
Please rebase on the latest master and it will be good to go.

PR #201, opened 2022-06-13
Make Kafka topics configurable

taktv6:
Making Kafka output topics configurable using an optional JSON config file. Topics that don't have a topic name assigned are ignored. This allows streaming only specific message types to Kafka and also makes it possible to run multiple GoBMP instances against the same Kafka cluster.

### DIFF
--- a/cmd/gobmp/gobmp.go
+++ b/cmd/gobmp/gobmp.go
@@ -1,8 +1,10 @@
 package main
 
 import (
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"os/signal"
 	"runtime"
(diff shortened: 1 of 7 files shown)


sbezverk:
@taktv6 please be advised, we have a relatively large installation base of gobmp already running, there are also other software modules which are importing and depending on topics names definitions. Removing topics constants will break those software modules. Therefore I will be accepting only changes which are not changing current behaviour, the changes must be optional and be disabled by default. Your desired behaviour should be turn on with explicit parameter.

taktv6:
Hi Serguei,
I think my patch fulfills all requirements you wrote. Messages are routed the very same way as they used to be. Once -config.file is specified the config is read and the output routed accordingly. Without a config file given the behavior doesn't change.
The topic name constants are not exported and thus can not be referenced by other modules. What am I missing here?

Cheers
takt

sbezverk:
@taktv6  We have a couple of use cases where ipv4 and ipv6 prefixes are going into the same collection. If you relay on IsIPv4 then this use case will be broken. If you restore `splitAF`, I think the rest (configurable topics) are great idea.

PR #204, opened 2022-09-21
fixed evpn route-type 5, ip-prefix, decoding

denizaydin:
EVPN ip-prefix route, route-type 5, is not properly decoded (RFC 9136). Added support for ip prefix route decoding for IPv4 and IPv6.


sbezverk (changes requested):
Please, add checks to make code safer. So the pointer `p` will never go above the length ob slice of bytes passed to Unmarshal func.
I would really appreciate if you could add a unit test where you process raw slice of bytes and then compare resulting object with what you expect.

sbezverk:
@denizaydin   Thank you very much for your PR, please address my comments above. Best regards..

denizaydin:
> Please, add checks to make code safer. So the pointer `p` will never go above the length ob slice of bytes passed to Unmarshal func.
> I would really appreciate if you could add a unit test where you process raw slice of bytes and then compare resulting object with what you expect.

hi @sbezverk,
theorically p can not go above the slice as we are calling different unmarshal functions for ipv4 or ipv6 prefixes. we comparing the length of the packet till the label. So if its ipv4, l=34 and if tis ipv6 l=58. This is total length till the label. and label unmarshalling goes to bottom label which always checks the if its hitting end of slice. Correct me if I am wrong.

I'll try to add unit tests.

denizaydin:
Hi @sbezverk,

I have checked the RFC again for evpn ip-prefix route. According the RFC9136, there can only be one label for ip-prefix routes unlike mpls vpn routes. So the loop that your are using for labels is not necassry thus the inital control of the unmarshalling evpn routes will do the exception control as we calling ipv4 and ipv6 according to its length, as the length is fixed there is no need to check the p pointer if inside unmarshall func as it already checked from calling func.

Could please send me some information how I am gonna use your testing for the code and how should I use. I can write go programs but not familiar with the testing side:( ( I am just a network engineer, knows some coding).  
By the way I am trying to set-up evpn datacenter bgp monitoring with using bmp, kafka and elk. I am at the begining of the project. I like to contribute on evpn side of your project.

sbezverk:
@denizaydin  Ok, in this case let's do the following, I will add necessary unit test, but I will need from you is RAW BGP update carrying evpn route. You need to run gobmp with `--v=6`, collect hex if the update and paste it here. Once I have it, I will add unit test and will be able to test your PR. I do not have evpn environment unfortunately.

denizaydin:
Hi @sbezverk,
I have added test for new prefix for ipv6 and ipv4. during testing I realized that you are not using correct masking for exp bit of mpls label. You may not hit the problem during your tests as the none of the test prefixes has exp bit > 0.
As my test setup is a vxlan evpn, mpls label points the vni not a common mpls header. I think of chaning label format but instead of doing that I just use as its common mpls label.
If we need the correct value for the label, I mean the vni, I will try to change the label structure to include a raw value composed of 24 bits.

Now evpn_test are successful but  I do not know why travis CI is failing? Can you help on that?

sbezverk:
@denizaydin  You need to run `gofmt` to format your source files. Usually it is done automatically if you use IDE. Otherwise  you need to do it manually.

pkg/evpn/evpn.go
```diff
@@ -131,9 +131,19 @@ func UnmarshalEVPNNLRI(b []byte) (*Route, error) {
 				return nil, err
 			}
 		case 5:
-			n.RouteTypeSpec, err = UnmarshalEVPNIPPrefix(b[p : p+l])
-			if err != nil {
-				return nil, err
+			switch l {
+				case 34:
+					n.RouteTypeSpec, err = UnmarshalEVPNIPv4Prefix(b[p : p+l])
```
sbezverk: These 2 Unmarshal func looks like do the same thing, please use  common  func for both and pass the prefix length as a parameter.

sbezverk (changes requested):
Please address the change  requested

sbezverk (approved):

sbezverk:
@denizaydin  Thank you, looks good.

PR #205, opened 2022-10-03
Couple of improvements for evpn

denizaydin:
- Added Peer-Type 3 (local-rib) RFC 9069.
- Added BMP statistic messages parsing.
- Updated Ext Community RMAC string represantation.
- Added some required fields to EVPN messages.

### DIFF
--- a/pkg/bgp/extended-community.go
+++ b/pkg/bgp/extended-community.go
@@ -286,15 +286,13 @@ func type6(subType uint8, value []byte) string {
 		copy(l, value[3:])
 		s = fmt.Sprintf("%d:%d", value[0], binary.BigEndian.Uint32(l))
 	case 0x02:
+		fallthrough
+	case 0x03:
 		for i, m := range value {
-			s += fmt.Sprintf("%02x", m)
+			s += tools.ConvertToHex(m)
 			if i < len(value)-1 {
 				s += ":"
 			}
-		}
-	case 0x03:
-		if len(value) == 6 {
-			s = fmt.Sprintf("%02x:%02x:%02x:%02x:%02x:%02x", value[0], value[1], value[2], value[3], value[4], value[5])
 		}
 	case 0x00:
 		s = fmt.Sprintf("%d:%d", value[0], binary.BigEndian.Uint32(value[2:]))
--- a/pkg/bgp/extended-community.go
+++ b/pkg/bgp/extended-community.go
@@ -248,7 +248,7 @@ func getSubType(m map[uint8]string, subType uint8) string {
 	return s
 }
 
-//Transitive Two-Octet AS-Specific Extended Community
+// Transitive Two-Octet AS-Specific Extended Community
 func type0(subType uint8, value []byte) string {
 	return getSubType(transAS2SubTypes, subType) + fmt.Sprintf("%d:%d", binary.BigEndian.Uint16(value[0:2]), binary.BigEndian.Uint32(value[2:]))
 }
@@ -258,7 +258,7 @@ func type1(subType uint8, value []byte) string {
 	return getSubType(transIPv4SubTypes, subType) + fmt.Sprintf("%s:%d", net.IP(value[0:4]).To4().String(), binary.BigEndian.Uint16(value[4:]))
 }
 
-//Transitive Four-Octet AS-Specific Extended Community
+// Transitive Four-Octet AS-Specific Extended Community
 func type2(subType uint8, value []byte) string {
 	return getSubType(transAS4SubTypes, subType) + fmt.Sprintf("%d:%d", binary.BigEndian.Uint32(value[0:4]), binary.BigEndian.Uint16(value[4:]))
 }
@@ -291,6 +291,10 @@ func type6(subType uint8, value []byte) string {
 			if i < len(value)-1 {
 				s += ":"
 			}
+		}
+	case 0x03:
+		if len(value) == 6 {
+			s = fmt.Sprintf("%02x:%02x:%02x:%02x:%02x:%02x", value[0], value[1], value[2], value[3], value[4], value[5])
 		}
 	case 0x00:
 		s = fmt.Sprintf("%d:%d", value[0], binary.BigEndian.Uint32(value[2:]))
--- a/pkg/bmp/per-peer-header.go
+++ b/pkg/bmp/per-peer-header.go
@@ -82,6 +82,7 @@ func UnmarshalPerPeerHeader(b []byte) (*PerPeerHeader, error) {
 	// *  Peer Type = 0: Global Instance Peer
 	// *  Peer Type = 1: RD Instance Peer
 	// *  Peer Type = 2: Local Instance Peer
+	// *  Peer Type = 3: Local RIB Peer, RFC9069
 	switch b[0] {
 	case 0:
 	case 1:
--- a/pkg/bmp/per-peer-header.go
+++ b/pkg/bmp/per-peer-header.go
@@ -86,6 +86,7 @@ func UnmarshalPerPeerHeader(b []byte) (*PerPeerHeader, error) {
 	case 0:
 	case 1:
 	case 2:
+	case 3:
 	default:
 		return nil, fmt.Errorf("invalid peer type, expected between 0 and 2 found %d", b[0])
 	}
--- a/pkg/evpn/evpn.go
+++ b/pkg/evpn/evpn.go
@@ -81,28 +81,8 @@ func (n *NLRI) GetEVPNGWAddr() []byte {
 }
 
 // GetEVPNLabel returns stack of labels found in the nlri
-func (n *NLRI) GetEVPNLabel() []uint32 {
-	label := make([]uint32, 0)
-	for _, l := range n.RouteTypeSpec.getLabel() {
-		label = append(label, l.Value)
-	}
-
-	return label
-}
-
-// GetEVPNRAWLabel returns stack of raw labels.
-// Raw label values is needed when encapsulation is not mpls, and label no longer is a standart mpls label.
-func (n *NLRI) GetEVPNRawLabel() []uint32 {
-	label := make([]uint32, 0)
-	for _, l := range n.RouteTypeSpec.getLabel() {
-		value := l.Value*16 + uint32(l.Exp*2)
-		if l.BoS {
-			value++
-		}
-		label = append(label, value)
-	}
-
-	return label
+func (n *NLRI) GetEVPNLabel() []*base.Label {
+	return n.RouteTypeSpec.getLabel()
 }
 
 // UnmarshalEVPNNLRI instantiates an EVPN NLRI object
--- a/pkg/evpn/evpn.go
+++ b/pkg/evpn/evpn.go
@@ -90,6 +90,21 @@ func (n *NLRI) GetEVPNLabel() []uint32 {
 	return label
 }
 
+// GetEVPNRAWLabel returns stack of raw labels.
+// Raw label values is needed when encapsulation is not mpls, and label no longer is a standart mpls label.
+func (n *NLRI) GetEVPNRawLabel() []uint32 {
+	label := make([]uint32, 0)
+	for _, l := range n.RouteTypeSpec.getLabel() {
+		value := l.Value*16 + uint32(l.Exp*2)
+		if l.BoS {
+			value++
+		}
+		label = append(label, value)
+	}
+
+	return label
+}
+
 // UnmarshalEVPNNLRI instantiates an EVPN NLRI object
 func UnmarshalEVPNNLRI(b []byte) (*Route, error) {
 	if glog.V(6) {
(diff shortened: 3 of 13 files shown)


pkg/bgp/extended-community.go
```diff
@@ -292,6 +292,10 @@ func type6(subType uint8, value []byte) string {
 				s += ":"
 			}
 		}
+	case 0x03:
```
sbezverk: It looks to me that processing for type 0x2 and 0x3 are the same. I would like to ask 2 things. 
For case:
case 0x02:
    fallthrough
case 0x03:
do the loop here for both cases.

instead of using `fmt.Sprintf` please use `tools.ConvertToHex()` it is tiny function will do processing much faster than heavy Sprintf. You can find it in `github.com/sbezverk/tools`

pkg/bmp/per-peer-header.go
```diff
@@ -86,6 +86,7 @@ func UnmarshalPerPeerHeader(b []byte) (*PerPeerHeader, error) {
 	case 0:
 	case 1:
 	case 2:
+	case 3:
```
sbezverk, line 90: Please add a comment after line 84 for a new Peer type 3

pkg/evpn/evpn.go
```diff
@@ -90,6 +90,21 @@ func (n *NLRI) GetEVPNLabel() []uint32 {
 	return label
 }
 
+// GetEVPNRAWLabel returns stack of raw labels.
```
sbezverk: It looks to me that it is not 100% EVPN specific function. Other protocols might be also interested in getting the raw label. Please extend `base.Label`  with new method Raw which would return uint32 of the raw label value. In this case computing of raw value will be in a single location and it does not need to be implemented by every NLRI which needs it,

pkg/kafka/kafka-publisher.go
```diff
@@ -34,6 +34,7 @@ const (
 	flowspecMessageTopic   = "gobmp.parsed.flowspec"
 	flowspecMessageV4Topic = "gobmp.parsed.flowspec_v4"
 	flowspecMessageV6Topic = "gobmp.parsed.flowspec_v6"
+	statsMessageTopic      = "gobmp.parsed.statictics"
```
sbezverk: typo should be `statistics` instead of `statictics`

pkg/message/evpn.go
```diff
@@ -75,15 +78,19 @@ func (p *producer) evpn(nlri bgp.MPNLRI, op int, ph *bmp.PerPeerHeader, update *
 				prfx.IPLength = *ip
 				gw := e.GetEVPNGWAddr()
 				// IPv4 should have IPLength set to 32
-				if prfx.IPLength == 32 {
-					prfx.IPAddress = net.IP(e.GetEVPNIPAddr()).To4().String()
+				if prfx.IPLength <= 32 {
+					if e.GetEVPNIPAddr() != nil {
```
sbezverk: no reason to execute `e.GetEVPNIPAddr()` twice.

if a := e.GetEVPNIPAddr(); a != nil {
   prfx.IPAddress = net.IP(a).To4().String()
}

Could you please comment or send a reference that IPLength can be less than 32, same comment to ipv6

pkg/message/types.go
```diff
@@ -381,3 +384,28 @@ type Flowspec struct {
...
+	Rev                        string `json:"_rev,omitempty"`
+	Sequence                   int    `json:"sequence,omitempty"`
+	RouterHash                 string `json:"router_hash,omitempty"`
+	RouterIP                   string `json:"router_ip,omitempty"`
+	RemoteBGPID                string `json:"remote_bgp_id,omitempty"`
+	RemoteASN                  uint32 `json:"remote_asn,omitempty"`
+	RemoteIP                   string `json:"remote_ip,omitempty"`
+	PeerRD                     string `json:"peer_rd,omitempty"`
+	Timestamp                  string `json:"timestamp,omitempty"`
+	DublicatePrefixs           uint32 `json:"dubprefixes,omitempty"`
```
sbezverk: typo in the variable name, please use `duplicate_prefix` in json, it is easier to read

pkg/message/types.go
```diff
@@ -381,3 +384,28 @@ type Flowspec struct {
...
+	DublicateWithDraws         uint32 `json:"dublicatewithdraws,omitempty"`
+	InvalidatedDueCluster      uint32 `json:"invalidatedduecluster,omitempty"`
+	InvalidatedDueAspath       uint32 `json:"invalidateddueaspath,omitempty"`
+	InvalidatedDueOriginatorId uint32 `json:"invalidateddueoriginatorid,omitempty"`
+	InvalidatedAsConfed        uint32 `json:"invalidateddueasconfed,omitempty"`
+	AdjRIBsIn                  uint64 `json:"adsribin,omitempty"`
+	LocalRib                   uint64 `json:"localrib,omitempty"`
+	UpdatesAsWithdraw          uint32 `json:"updatesaswithdraw,omitempty"`
+	PrefixesAsWithdraw         uint32 `json:"prefixesaswithdraw,omitempty"`
```
sbezverk: for all compound names use `blah_blah` format in json

sbezverk (changes requested):
please review requested changes and let me know if you have questions.

pkg/message/evpn.go
denizaydin: For evpn route-type 2, MAC-IP route type it's 32.
For ip prefix type, type-5 it may be anything a default route with 0 or /24 route e.t.c.
The problem is route type 2 and 5 uses the same object for storing the ip prefix. I do not find a reason two use different structure, your approach seems much more simpler.

pkg/evpn/evpn.go
denizaydin: From programing point of view best thing is chaning label structure to inclue only raw label and assign methods to get exp and bos bit like .GetExp and GetBos. But I gues this will require more change and and cause trouble where they are used. Although I do not see the use of exp and bos bit in the code. Nevertheless it will requrire lots of change on test methods:)) So I will add GetRawValue which will be a little bit ugly:)

denizaydin: dont we still need a method on evpn/evpn.go to get raw value from messages/evpn.go? There is no method returning the label itself on evpn/evpn.go. Or I can not figure it out. Do you have anything in your mind?

denizaydin: Ok figured it out. But this need change evpn/evpn.go GetEVPNLabel method to return labels itself.

denizaydin:
Should we add peer type ( global, local, local-rib) to all message types?

pkg/bgp/extended-community.go
```diff
@@ -286,8 +286,10 @@ func type6(subType uint8, value []byte) string {
 		copy(l, value[3:])
 		s = fmt.Sprintf("%d:%d", value[0], binary.BigEndian.Uint32(l))
 	case 0x02:
+		fallthrough
+	case 0x03:
```
sbezverk, line 290: Should it be 0x5?? I think in the previous commit it was 0x5 if I am not mistaken.

pkg/message/types.go
```diff
@@ -381,3 +384,28 @@ type Flowspec struct {
...
+	Rev                        string `json:"_rev,omitempty"`
+	Sequence                   int    `json:"sequence,omitempty"`
+	RouterHash                 string `json:"router_hash,omitempty"`
+	RouterIP                   string `json:"router_ip,omitempty"`
+	RemoteBGPID                string `json:"remote_bgp_id,omitempty"`
+	RemoteASN                  uint32 `json:"remote_asn,omitempty"`
+	RemoteIP                   string `json:"remote_ip,omitempty"`
+	PeerRD                     string `json:"peer_rd,omitempty"`
+	Timestamp                  string `json:"timestamp,omitempty"`
+	DublicatePrefixs           uint32 `json:"dublicate_prefix,omitempty"`
```
sbezverk: Please fix var name and json tag

pkg/message/types.go
```diff
@@ -381,3 +384,28 @@ type Flowspec struct {
...
+	DublicateWithDraws         uint32 `json:"dublicate_withdraws,omitempty"`
```
sbezverk: Please fix var name and json tag

sbezverk (changes requested):
Almost there, a couple of nits to address.

pkg/bgp/extended-community.go
denizaydin: Nope, it was. 0X03, EVPN Router’s MAC Extended Community | [RFC9135]

denizaydin:
is there anything that needs to be addressed?

sbezverk:
@denizaydin  I will need to run this change against bgp router supporting local-rib and peer type 3. I should be done within next few days, once it is successfully tested, this PR will be merged.

sbezverk:
@denizaydin 
I run gobmp with you PR, as soon as bmp peer with enabled local-rib comes up, I see this panic:
```
I1012 05:01:09.187692       1 gobmpsrv.go:52] client 10.1.1.100:4732 accepted, calling bmpWorker
panic: runtime error: slice bounds out of range [202:10]

goroutine 64 [running]:
github.com/sbezverk/gobmp/pkg/bmp.UnmarshalPeerUpMessage({0xc0001ae030, 0xd4, 0xd4})
        /Users/sbezverk/projects/go/workspace/gobmp/pkg/bmp/peer-up.go:61 +0x4b2
github.com/sbezverk/gobmp/pkg/parser.parsingWorker({0xc0001ae000, 0x104, 0x104}, 0xc00006afc0)
        /Users/sbezverk/projects/go/workspace/gobmp/pkg/parser/parser.go:81 +0x385
created by github.com/sbezverk/gobmp/pkg/parser.Parser
        /Users/sbezverk/projects/go/workspace/gobmp/pkg/parser/parser.go:14 +0x47
```

Here is the output of PeerUp Raw message causing the crash
```
I1012 05:06:59.303434       1 peer-up.go:24] BMP Peer Up Message Raw: [ 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x5B, 0x01, 0x04, 0xC3, 0xCB, 0x00, 0x00, 0x01, 0x01, 0x0A, 0x01, 0x3E, 0x02, 0x06, 0x01, 0x04, 0x00, 0x01, 0x00, 0x01, 0x02, 0x06, 0x01, 0x04, 0x00, 0x01, 0x00, 0x04, 0x02, 0x06, 0x01, 0x04, 0x00, 0x01, 0x00, 0x80, 0x02, 0x02, 0x80, 0x00, 0x02, 0x02, 0x02, 0x00, 0x02, 0x06, 0x41, 0x04, 0x00, 0x00, 0xC3, 0xCB, 0x02, 0x14, 0x05, 0x12, 0x00, 0x01, 0x00, 0x01, 0x00, 0x02, 0x00, 0x01, 0x00, 0x02, 0x00, 0x02, 0x00, 0x01, 0x00, 0x80, 0x00, 0x02, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0x00, 0x5B, 0x01, 0x04, 0xC3, 0xCB, 0x00, 0x00, 0x01, 0x01, 0x0A, 0x01, 0x3E, 0x02, 0x06, 0x01, 0x04, 0x00, 0x01, 0x00, 0x01, 0x02, 0x06, 0x01, 0x04, 0x00, 0x01, 0x00, 0x04, 0x02, 0x06, 0x01, 0x04, 0x00, 0x01, 0x00, 0x80, 0x02, 0x02, 0x80, 0x00, 0x02, 0x02, 0x02, 0x00, 0x02, 0x06, 0x41, 0x04, 0x00, 0x00, 0xC3, 0xCB, 0x02, 0x14, 0x05, 0x12, 0x00, 0x01, 0x00, 0x01, 0x00, 0x02, 0x00, 0x01, 0x00, 0x02, 0x00, 0x02, 0x00, 0x01, 0x00, 0x80, 0x00, 0x02, 0x00, 0x03, 0x00, 0x06, 0x67, 0x6C, 0x6F, 0x62, 0x61, 0x6C ]
```
please investigate this panic.

denizaydin:
@sbezverk the problem is not about the PR, its was already a bug.
 -  peer.go (line 61) -> **tlvs, err := UnmarshalTLV(b[p : len(b)-int(p)])** should be tlvs, err := UnmarshalTLV(b[p:]).
 
Should I change it?

sbezverk:
@denizaydin you are right, please do.

sbezverk (approved):
Thank you for PR, approved

PR #207, opened 2022-10-13
 Deal with different sets of flags for different Peer types in PerPeerHeader

sbezverk:
With introduction of PeerType 3 in PerPeerHeader, the possible set of flags has changed, Peer type 0,1 and 2 expects VLAO, flags where as Peer type 3 expects only F flag. This PR provides methods to PerPeerHeader to deal with flags consist to the type.


sbezverk:
@brmcdoug Appreciate if you could give this image ` docker.io/sbezverk/gobmp:wip_10_13_2022` a spin in your lab and let me know the observations. 
@denizaydin  Please test it as well, to confirm it does not break for you on `evpn` side.

denizaydin:
@sbezverk . I can not see any error.
I am using goBGP for relaying routes from ToR switches/Spines to the BMP server. Local RIB mimics the spine route selection on the goBGP. Although BMP RFC defines a type of peer-up messages for local-rib, I am not able to replicate  with goBGP. So I am can not able to test peer-up for local-rib. Other than that it seems fine.

sbezverk:
@denizaydin  Thank you, for confirmation. I was able to test Peer Type 3, the support  has just been introduced in Cisco XR routers.

denizaydin:
@sbezverk I am also woring on peer-down, waiting for your PR than gonna push it. It contains new type 6 for vrf delete or e.t.c event for local rib. 
I see that you are already processing this information via _information tlv_ in peer-up.go howevet this information is not included in peer messages.

PR #212, opened 2023-01-13
Implement BMP RAW message type

tobez:
This is an implementation of BMP RAW messages that can be found in OpenBMP (as described in https://github.com/SNAS/openbmp/blob/master/docs/MESSAGE_BUS_API.md).

Production of the raw messages can be switched on by specifying --bmp-raw on the command line.
Additionally, a --admin-id=STRING option is added in order to have a control over collector hash ID mentioned in the RAW format specification.

Specifying --bmp-raw only makes sense in case Kafka publishing is in effect.

The implementation is fairly straightforward as the producer needs to do less work than in the case of parsed messages produced by default.

This work is based on an E-mail discussion between goBMP's author Serguei Bezverkhi @sbezverk and David Teach @dteach-rv, and was sponsored by NSRC https://nsrc.org/

### DIFF
--- a/pkg/gobmpsrv/gobmpsrv.go
+++ b/pkg/gobmpsrv/gobmpsrv.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"os"
 
 	"github.com/golang/glog"
 	"github.com/sbezverk/gobmp/pkg/bmp"
@@ -27,6 +28,7 @@ type bmpServer struct {
 	destinationPort int
 	incoming        net.Listener
 	stop            chan struct{}
+	adminId         string
 }
 
 func (srv *bmpServer) Start() {
@@ -69,7 +71,7 @@ func (srv *bmpServer) bmpWorker(client net.Conn) {
 		glog.V(5).Infof("connection to destination server %v established, start intercepting", server.RemoteAddr())
 	}
 	var producerQueue chan bmp.Message
-	prod := message.NewProducer(srv.publisher, srv.splitAF)
+	prod := message.NewProducer(srv.publisher, srv.adminId, srv.splitAF)
 	prodStop := make(chan struct{})
 	producerQueue = make(chan bmp.Message)
 	// Starting messages producer per client with dedicated work queue
@@ -134,6 +136,14 @@ func NewBMPServer(sPort, dPort int, intercept bool, p pub.Publisher, splitAF boo
 		splitAF:         splitAF,
 		bmpRaw:          bmpRaw,
 	}
+
+	// XXX need to be able specify admin id via a command line switch
+	// But for now, just use the hostname
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "dummy"
+	}
+	bmp.adminId = hostname
 
 	return &bmp, nil
 }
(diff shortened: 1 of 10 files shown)


pkg/gobmpsrv/gobmpsrv.go
```diff
@@ -68,7 +70,7 @@ func (srv *bmpServer) bmpWorker(client net.Conn) {
 		glog.V(5).Infof("connection to destination server %v established, start intercepting", server.RemoteAddr())
 	}
 	var producerQueue chan bmp.Message
-	prod := message.NewProducer(srv.publisher, srv.splitAF)
+	prod := message.NewProducer(srv.publisher, srv.adminId, srv.splitAF)
```
sbezverk, line 73: Since admin ID is somewhat publisher requirement, it is better to keep in in a publisher. You can add a method to publisher something like getAdminId and call it when message gets published adding it to the produced message.

pkg/gobmpsrv/gobmpsrv.go
```diff
@@ -77,7 +79,7 @@ func (srv *bmpServer) bmpWorker(client net.Conn) {
 	parserQueue := make(chan []byte)
 	parsStop := make(chan struct{})
 	// Starting parser per client with dedicated work queue
-	go parser.Parser(parserQueue, producerQueue, parsStop)
+	go parser.Parser(parserQueue, producerQueue, parsStop, srv.bmpRaw)
```
sbezverk, line 82: not great idea changing this function parameters with a one publisher specific parameter, please find a way to pass this parameter by other means

pkg/gobmpsrv/gobmpsrv.go
```diff
@@ -117,7 +119,7 @@ func (srv *bmpServer) bmpWorker(client net.Conn) {
 }
 
 // NewBMPServer instantiates a new instance of BMP Server
-func NewBMPServer(sPort, dPort int, intercept bool, p pub.Publisher, splitAF bool) (BMPServer, error) {
+func NewBMPServer(sPort, dPort int, intercept bool, p pub.Publisher, splitAF bool, bmpRaw bool, adminId string) (BMPServer, error) {
```
sbezverk, line 122: same comment as above, one publisher specific parameters should not change this function signature. Try to pass this information via publisher interface.

pkg/message/raw.go
```diff
@@ -0,0 +1,32 @@
...
+		glog.Errorf("perPeerHeader is missing, cannot construct proper raw message")
+		return
+	}
+	rm, ok := msg.Payload.(*bmp.RawMessage)
+	if !ok {
+		glog.Errorf("got invalid Payload type in bmp.RawMessage %+v", msg.Payload)
+		return
+	}
+
+	out := []byte(fmt.Sprintf("V: 1.7\nC_HASH_ID: %s\nR_HASH: %s\nR_IP: %s\nL: %d\n\n",
```
sbezverk, line 23: Sprintf in golang is very costly,  please try to optimize the process of building the message.

sbezverk (changes requested):
If you are still interested in completing this functionality, let's work on making some changes which would not extend functions prototypes, but passing desired setting via interfaces' calls.

sbezverk:
Closed as implemented in #283

Issue #215, opened 2023-03-20
bgpls.GetUnreservedLinkBandwidthKbps is appending initial values but should be replacing the zero values


slinkish:
I noticed that data specific to `unresv_bw_kbps` was being produced as a list that is too long for the actual TE TLV parameters. For example,
```
"unresv_bw_kbps":[0,0,0,0,0,0,0,0,20000000,20000000,20000000,20000000,20000000,20000000,20000000,20000000]
```
Notice there are 8 0 values and then 8 true values from the device. Upon investigation, I found bgpls.GetUnreservedLinkBandwidthKbps is initializing and empty slice of length 8 and then this slice is being appended with the actual TLV data producing the slice of length 16. 

I have a PR that I will submit shortly that rejiggers the for loop logic to index the initial slice so that the original 0's get replaced with TLV specific data and the final slice is of length 8 as expected. 
Expected result:
```
"unresv_bw_kbps":[20000000,20000000,20000000,20000000,20000000,20000000,20000000,20000000]
```

slinkish:
See https://github.com/sbezverk/gobmp/pull/216

PR #216, opened 2023-03-20
Index unresv bw slice

slinkish:
See https://github.com/sbezverk/gobmp/issues/215

### DIFF
--- a/pkg/bgpls/bgpls-nlri.go
+++ b/pkg/bgpls/bgpls-nlri.go
@@ -460,9 +460,8 @@ func (ls *NLRI) GetUnreservedLinkBandwidthKbps() []uint64 {
 		if tlv.Type != 1091 {
 			continue
 		}
-		for p := 0; p < len(tlv.Value); {
-			unResrved = append(unResrved, uint64(math.Float32frombits(binary.BigEndian.Uint32(tlv.Value[p:p+4])) * 8 / 1000))
-			p += 4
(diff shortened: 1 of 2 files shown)


pkg/bgpls/bgpls-nlri.go
```diff
@@ -460,9 +460,8 @@ func (ls *NLRI) GetUnreservedLinkBandwidthKbps() []uint64 {
 		if tlv.Type != 1091 {
 			continue
 		}
-		for p := 0; p < len(tlv.Value); {
-			unResrved = append(unResrved, uint64(math.Float32frombits(binary.BigEndian.Uint32(tlv.Value[p:p+4])) * 8 / 1000))
-			p += 4
+		for i, p := 0, 0; p < len(tlv.Value); i, p = i+1, p+4 {
+			unResrved[i] = uint64(math.Float32frombits(binary.BigEndian.Uint32(tlv.Value[p:p+4])) * 8 / 1000)
```
sbezverk, line 469: the issue exists in the old code too, in case tlv length is wrong, unchecked p:p+4 will cause a panic. Do you want to add extra length check just to be on the safe side?

slinkish: Yeah sounds good, I added commit that checks the length, in the case that TLV is not expected length, function will return the initial value of all zeros so as to not introduce new error handling but does log the error so this event should not be too horribly obfuscated.

sbezverk (approved):

Issue #220, opened 2023-05-12
Support initiating outbound BMP connections


jarrodb:
Various vendors support defining passive BMP connections that allow the client to initiate the connection.  Support defining outbound connections in a backwards compatible way.

sbezverk:
Is this supported by some BMP related  RFC or a proposal? I do not remember seeing that in RFC. Please provide a reference in the existing or new RFC, otherwise I would prefer to stay within the standard.

jarrodb:
This support would be 🔥 since it wouldn't require static station configurations and would allow for gobmp collectors to be moved around per whatever container orchestration is being deployed as long as CoPP supported the connection.

sbezverk:
I am not so sure about that. We had some interesting feedback from some major service providers, that they liked that BMP is one way communication, the collector never initiates a connection or needlessly poll routers for connections.  Only routers, when they are ready/configured would establish BMP connection and start pushing information.  I will have to run again the reference to RFC you kindly provided to see what they think about it.

jarrodb:
I definitely wasn't recommending making that the default behavior, but all router vendors support passive connections per the RFC and I think there is a valid use case.  I would be hopeful that we could support that option in this project via an option flag.

sbezverk:
Please let me know BGP vendor who supports this functionality, so far I could not find any, also feedback from the service providers is negative with regards to "security implications" with this type of connectivity when the collector is active and a bgp instance is passive.

jarrodb:
The two that comprise the majority of our network:

[Juniper BMP](https://www.juniper.net/documentation/us/en/software/junos/bgp/topics/ref/statement/bmp-edit-routing-options.html)
```
bmp {
    station station-name {
        connection-mode (active | passive);
```

and

[Arista BMP](https://www.arista.com/en/support/toi/eos-4-20-1f/13924-bgp-monitoring-protocol)
```
router bgp 64512
   monitoring station BMP1
     connection mode passive
     connection address 192.168.1.5
```

I really haven't looked at others yet.

jarrodb:
Also, I'm not sure I understand the security implication reasoning.  Any listening socket is a potential security implication if enabled on a router without proper filtering and other security policies.  However, supporting outbound connections at the collector level doesnt force a service provider to enable it on their router(s) since it's a feature flag that is only enabled if desired.

jarrodb:
[...]

PR #223, opened 2023-06-21
Initial passive BMP support

jarrodb:
This PR is an initial iteration to support passive outbound BMP connections as a backwards-compatible _option_ when initializing the BMP server (see #220 for reference).  As stated in the issue thread, the RFC defines support for passive connections and is supported by major routing device vendors (e.g. Arista and Juniper).

The initial approach is simplistic, but my plan would be to support for multiple routers to establish multiple connections in parallel or failover schemas.

### DIFF
--- a/cmd/gobmp/gobmp.go
+++ b/cmd/gobmp/gobmp.go
@@ -22,15 +22,16 @@ import (
 )
 
 var (
-	dstPort   int
-	srcPort   int
-	perfPort  int
-	kafkaSrv  string
-	natsSrv   string
-	intercept string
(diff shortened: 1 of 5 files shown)


jarrodb:
Gunna need to add authentication next..

sbezverk:
Based on the code so far, in a pssive mode, the connection is initiated to a single router, unless I miss where a list of BGP instances is handled. One gobmp per passivle bgp speaker is not a efficient approach. Closing it for now, but feel free to reopen and start discussion.

PR #230, opened 2023-08-19
fix panic in nlri 71 and add unit tests

sbezverk:
PR fixes panic in nlri 71 processing, when nlri 71 carries a malformed message


sbezverk:
closing it as outdated

PR #244, opened 2024-10-14
added configurable Kafka retention time

umkhan123:
added configurable 
kafkaTpRetnTimeMs, "kafka-topic-retention-time-ms", "900000", "Kafka topic retention time in ms, default is 900000 ms i.e 15 minutes")


sbezverk (changes requested):
Please add value validation logic to be in a sane range. Also introduce a kafka configuration struct, if in future other parameters need to be specified, parameters list of NewKafkaPublisher will not grow indefinitely.

umkhan123:
added the validation for retention time and also Kafka config struct.

pkg/kafka/kafka-publisher.go
```diff
@@ -202,8 +200,8 @@ func NewKafkaPublisher(kafkaSrv string) (pub.Publisher, error) {
 	}, nil
 }
 
-func validator(addr string) error {
-	host, port, _ := net.SplitHostPort(addr)
+func validator(kConfig Config) error {
```
sbezverk: Please pass kConfig by reference and check for nil to prevent panic when dereferencing..

pkg/kafka/kafka-publisher.go
```diff
@@ -221,17 +219,25 @@ func validator(addr string) error {
 	if np == 0 || np > math.MaxUint16 {
 		return fmt.Errorf("the value of port is invalid")
 	}
+	// validator for topic retention time
+	if i, err := strconv.Atoi(kConfig.TopicRetentionTimeMs); err != nil {
+		return err
+	} else {
```
sbezverk: else is not needed, as in case of err != nil the function will return.

pkg/kafka/kafka-publisher.go
```diff
@@ -221,17 +219,25 @@ func validator(addr string) error {
...
+		if i < 0 {
```
sbezverk: There is one case when retention can be set to -1, (forever) probably for completeness worth to integrate.

pkg/kafka/kafka-publisher.go
```diff
@@ -221,17 +219,25 @@ func validator(addr string) error {
...
+			return fmt.Errorf("kafka topic retention time can not be less than 0")
+		}
+	}
 	return nil
 }
 
-func ensureTopic(br *sarama.Broker, timeout time.Duration, topicName string) error {
+func ensureTopic(br *sarama.Broker, timeout time.Duration, topicName, kafkaTpRetnTimeMs string) error {
```
sbezverk: Please use kafka config struct

pkg/kafka/kafka-publisher.go
```diff
@@ -144,10 +142,10 @@ func (p *publisher) Stop() {
 }
 
 // NewKafkaPublisher instantiates a new instance of a Kafka publisher
-func NewKafkaPublisher(kafkaSrv string) (pub.Publisher, error) {
+func NewKafkaPublisher(kConfig *Config) (pub.Publisher, error) {
```
sbezverk, line 145: check for nil

sbezverk (changes requested):
Almost there, few minor changes and good to go..

pkg/kafka/kafka-publisher.go
umkhan123: this is now checked in validator on line 147 .

umkhan123:
@sbezverk can you please review

sbezverk (approved):

sbezverk:
thank you

PR #246, opened 2024-11-20
Add timeout to create topic request in Kafka publisher

### LINKED ISSUE #210: createTopic timeouts
ensureTopic method does not have an actual timeout for topic creation, and it seems default value for this is "0", yes an actual zero.  So it fails to create topics at least in our environment

adding `		Timeout: timeout,` at line 236 in kafka-publisher.go worked for us.

klegul:
This fixes issues with some Kafka environments and closes #210.


sbezverk (approved):

sbezverk:
thank you

PR #256, opened 2025-03-17
Fixed logic for l3-vpn withdraw prefixes causing crash

iskhan-eqix:
**Issue faced:**
When L3VPN routes are withdrawn the prefix length comes up as 118bits leaving 30bits for IPv4 preifx after accounting for label stack(or compatibility field 3bytes) and the 8-byte RD. The fixed overhead is 3x8+8x8=24+64 = 88bits and this leaves 30bits(i.e., 118-88=30) for the IPv4 which is not full 32bits. The current logic seems to be computing NLRI total bytes using a ceiling division then subtracts the label and RD bytes, this leads to extracting 4bytes as the prefix, and then it sets the route prefix length to 32bits. This has been causing crash to the collector and post optimization it is stabilized and tested.
 
withdrawn-snap-pcap-ts

**Few Errors leading to crash:**
**Pre-fix some errs**

E0223 18:55:59.831667  664302 rd.go:27] MakeRD: invalid rd type detected in [ 0x84, 0x06, 0xC3, 0x50, 0xC8, 0x00, 0x02, 0x70 ]
E0223 18:55:59.831678  664302 rd.go:27] MakeRD: invalid rd type detected in [ 0x84, 0x06, 0xC3, 0x50, 0xC8, 0x00, 0x02, 0x70 ]
...
...

github.com/sbezverk/gobmp/pkg/bgp.(*MPUnReachNLRI).GetNLRIL3VPN(0xc000226090)
        /home/ish/i-gobmp/gobmp/pkg/bgp/mp-unreach-nlri.go:79 +0x6e fp=0xc074e73a18 sp=0xc074e739d0 pc=0x71c4ee
github.com/sbezverk/gobmp/pkg/message.(*producer).l3vpn(0xc0002b7880, {0x9db010, 0xc000226090}, 0x1, 0xc00016c000, 0xc000144d80)
        /home/ish/i-gobmp/gobmp/pkg/message/l3-vpn.go:14 +0x56 fp=0xc074e73c88 sp=0xc074e73a18 pc=0x722276
github.com/sbezverk/gobmp/pkg/message.(*producer).processMPUpdate(0xc0002b7880, {0x9db010, 0xc000226090}, 0x1, 0xc00016c000, 0xc000144d80)
        /home/ish/i-gobmp/gobmp/pkg/message/process-mp-update.go:63 +0x174 fp=0xc074e73e18 sp=0xc074e73c88 pc=0x7277f4
github.com/sbezverk/gobmp/pkg/message.(*producer).produceRouteMonitorMessage(0xc0002b7880, {0xc00016c000?, {0x8549c0?, 0xc00012eb30?}})
        /home/ish/i-gobmp/gobmp/pkg/message/route-monitor.go:56 +0x285 fp=0xc074e73f48 sp=0xc074e73e18 pc=0x729885
github.com/sbezverk/gobmp/pkg/message.(*producer).producingWorker(0xc00059ffdb?, {0xc00016c000?, {0x8549c0?, 0xc00012eb30?}})
        /home/ish/i-gobmp/gobmp/pkg/message/producer.go:48 +0xd9 fp=0xc074e73fb0 sp=0xc074e73f48 pc=0x729539
github.com/sbezverk/gobmp/pkg/message.(*producer).Producer.gowrap1()
        /home/ish/i-gobmp/gobmp/pkg/message/producer.go:33 +0x2c fp=0xc074e73fe0 sp=0xc074e73fb0 pc=0x72942c
runtime.goexit({})
        /usr/local/go/src/runtime/asm_amd64.s:1700 +0x1 fp=0xc074e73fe8 sp=0xc074e73fe0 pc=0x478901
created by github.com/sbezverk/gobmp/pkg/message.(*producer).Producer in goroutine 114
        /home/ish/i-gobmp/gobmp/pkg/message/producer.go:33 +0x4e

**Post-fix**

Graceful handling of withdrawn l3vpn prefix(es)..

E0227 19:07:17.446428 1069651 process-mp-update.go:65] failed to produce l3vpn messages with error: NLRI length is 0
E0227 19:07:17.446487 1069651 process-mp-update.go:65] failed to produce l3vpn messages with error: not found
E0227 19:07:17.446525 1069651 process-mp-update.go:85] failed to produce evpn messages with error: NLRI length is 0
E0227 19:48:47.209091 1069651 process-mp-update.go:65] failed to produce l3vpn messages with error: not enough bytes for prefix
E0227 19:52:40.125055 1069651 process-mp-update.go:65] failed to produce l3vpn messages with error: invalid NLRI length (0 bits)
E0227 19:52:40.125152 1069651 process-mp-update.go:65] failed to produce l3vpn messages with error: invalid NLRI length (0 bits)
E0227 19:53:14.691586 1069651 process-mp-update.go:65] failed to produce l3vpn messages with error: not enough bytes for prefix
E0227 19:53:14.691649 1069651 process-mp-update.go:65] failed to produce l3vpn messages with error: not enough bytes for prefix

**Summary of fixes:**

This fix adjusts the calculation so that it correctly derives the 30-bit prefix from 15-byte NLRI data rather than assuming 32-bits. 

Instead of using _int(up.Length/8)_ ; which truncates the up.Length is not a multiple of 8, then conditionally adding one, we compute,

		prefixBytes := (computedPrefixBits + 7) / 8
		if p+prefixBytes > len(b) {
			return mpnlriOrErr(&mpnlri, fmt.Errorf("not enough bytes for prefix"))
		}
		up.Prefix = make([]byte, prefixBytes)
		copy(up.Prefix, b[p:p+prefixBytes])
		p += prefixBytes

		// Set the route's prefix length
		up.Length = uint8(computedPrefixBits)

		mpnlri.NLRI = append(mpnlri.NLRI, up)

So if computed bits are < 32 bits we force to 32bits.

- Parses the L3VPN NLRI (including withdrawn routes) with the improved logic now (forced 32-bit prefix if needed, default RD if errors, etc.)
- Attempts a single fallback with !pathID if the first parse fails—without recursing infinitely
- Default RD if not enough bytes or if MakeRD fails
- Label logic handles both “no label” markers (0x80,0x00,0x00 or 0x00,0x00,0x00) and actual labels
- Avoids stack overflows by not calling UnmarshalL3VPNNLRI from inside itself on every error. No recursion or repeated error logs that could blow up the stack

### DIFF
--- a/pkg/l3vpn/l3-vpn.go
+++ b/pkg/l3vpn/l3-vpn.go
@@ -4,20 +4,15 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
-	"net"
 
 	"github.com/golang/glog"
 	"github.com/sbezverk/gobmp/pkg/base"
 	"github.com/sbezverk/tools"
 )
 
-// UnmarshalL3VPNNLRI tries to parse the L3VPN NLRI once using the given pathID, and if that fails, it tries again
-// once with !pathID. This avoids infinite recursion and prevents stack overflows...
+// UnmarshalL3VPNNLRI parses VPNv4/VPNv6 NLRI according to the caller-provided pathID flag
+// If parsing fails, it will try exactly once with !pathID. No recursion...
 func UnmarshalL3VPNNLRI(b []byte, pathID bool, srv6 ...bool) (*base.MPNLRI, error) {
-	if len(b) == 0 {
-		// Return empty NLRI structure instead of error
-		return &base.MPNLRI{NLRI: make([]base.Route, 0)}, nil
-	}
 	srv6Flag := false
 	if len(srv6) == 1 {
 		srv6Flag = srv6[0]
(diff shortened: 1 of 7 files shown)


sbezverk:
@iskhan-eqix  thank you for looking into this issue. As you see some unit tests are failing and they are testing different aspects of label handling. l3vpn processing is very critical so all proposed changes must be validated. Could you please run gobmp with --v=9 and provide me bgp update/withdraw message causing the crash.

pkg/l3vpn/l3-vpn.go
```diff
@@ -10,124 +10,157 @@ import (
...
+
+		// Next 3 bytes: label or compatibility field
 		if p+3 > len(b) {
-			err = fmt.Errorf("not enough bytes to reconstruct l3vpn nlri")
-			goto error_handle
+			return mpnlriOrErr(&mpnlri, fmt.Errorf("not enough bytes for label field"))
 		}
-		if bytes.Equal([]byte{0x80, 0x00, 0x00}, b[p:p+3]) {
+		labelField := b[p : p+3]
+		if bytes.Equal(labelField, []byte{0x80, 0x00, 0x00}) || bytes.Equal(labelField, []byte{0x00, 0x00, 0x00}) {
```
sbezverk: why ignore Explicit Null label?

iskhan-eqix: Explicit null needs to have BoS set in the final byte e.g., 00 00 01= BoS set to enabled. But 0x00 has no bit set for BoS. For withdrawn L3VPN routes a 3-byte sequence of all zeros (0x00, 0x00, 0x00) is simply used as a "no‐label" or compatibility marker.. The same applies to 0x80, 0x00, 0x00 its not a valid label, its a vendor‐specific indication that no label stack is carried. So we treat both (0x80, 0x00, 0x00 and 0x00, 0x00, 0x00) as no label present rather than an explicit null label. An actual MPLS label field explicit null or otherwise would set the BoS bit and/or carry a non‐zero label value in the top 20 bits. So if we see a plain 0x00, 0x00, 0x00 with no BoS bit set  we know it is not a valid label, hence we skip that compatibility marker indicating no label...

iskhan-eqix:
> @iskhan-eqix thank you for looking into this issue. As you see some unit tests are failing and they are testing different aspects of label handling. l3vpn processing is very critical so all proposed changes must be validated. Could you please run gobmp with --v=9 and provide me bgp update/withdraw message causing the crash. Message that starts with bgp marker.

Hi @sbezverk thanks for the review and test, I actually missed unicast.go mod - would be great if you can re-run tests. I will setup the original repo again and provide you with debugs as well. Thanks

sbezverk (approved):
third party testing has been successfully completed

Issue #261, opened 2025-06-19
Potential mis-ordering when parsing messages from a client and producing them on kafka?


reshad-equinix:
The parser's `Parser()` function runs in a goroutine as shown below: https://github.com/sbezverk/gobmp/blob/9fbd0c3830ec74452a2e56351efa372b086808d7/pkg/gobmpsrv/gobmpsrv.go#L79-L80

And messages read from the client are sent to the parser via the parserQueue channel:
https://github.com/sbezverk/gobmp/blob/9fbd0c3830ec74452a2e56351efa372b086808d7/pkg/gobmpsrv/gobmpsrv.go#L115

And for every message received, the parser does its work in a goroutine:
https://github.com/sbezverk/gobmp/blob/9fbd0c3830ec74452a2e56351efa372b086808d7/pkg/parser/parser.go#L10-L14

At the end of its work the parser sends the parsed BMP message to the producer via the producerQueue:
https://github.com/sbezverk/gobmp/blob/9fbd0c3830ec74452a2e56351efa372b086808d7/pkg/parser/parser.go#L105-L107

So it seems possible that 2 back-to-back messages from a client device would be handled in parallel because of `go parsingWorker(msg, producerQueue)` in the `Parser()` function. And theoretically add/del messages could get handled in an incorrect order by the producer. Unless I've misunderstood this part of the code :-)

It seems overkill to have multiple goroutines to process messages from the same client. I'd make a synchronous call to parsingWorker() instead. But I don't have the history...
Update: I just clued in that it's to handle the `stop` msg/chan. But I think it's an issue as mentioned above. Use WaitGroup?

reshad-equinix:
I've managed to reproduce the misordering issue via UT, and fix seems to be working. I'll open a PR "soon".

PR #262, opened 2025-06-20
Address potential parser misordering issue by using a WaitGroup.

reshad-equinix:
- Added a WaitGroup so that a call to parsingWorker() starts only if the previous call has completed.
- Also added a unit-test to catch misordering. This test fails if we remove the call to `wg.Wait()` in Parser`()`

https://github.com/sbezverk/gobmp/issues/261


sbezverk:
@reshad-equinix  in the workflow go version needs to be bumped us to support toolchain in go.mod.

reshad-equinix:
> @reshad-equinix the goal was to make gobmp to accept messages for the processing as fast as possible to avoid any queuing on a router's side. The order of messages should be dealt by the consumer as each message carries a timestamp. This race is possible when artificially created, but in the real life such condition is very unlikely as router would need to produce Add and Remove within milliseconds apart. Possible, but highly unlikely. 

I agree that it's unlikely but if it happens the consumer has incorrect state forever until the "route" in question changes again. And it could happen not just based on router behaviour, but also if there's delays in the network or if the BMP client was delayed reading or .... And if it happens there's no way to happen whether it actually happened, that's the part which worries me too.

As for putting the responsibility of ordering on the consumer, I find that onerous. The data stream is coming in "ordered" and IMO it's acceptable for some consumers to expect the data to remain ordered. 

> If you add serialization of processing, adding waitgroup seems an overkill. The easiest serialization would be just making parsingWorker to be a blocked call.
> 

I agree, that was actually my first intention but for some misguided reason (that I can't remember anymore) I thought that removing the async call wouldn't work :-(

> Please try to use this approach and test it on an internet feed.

Performance will suffer, by how much I'm not sure.
What if I make the synchronous behaviour (to guarantee ordering) enabled via config and the default behaviour stays as-is today?

reshad-equinix:
> @reshad-equinix in the workflow go version needs to be bumped us to support toolchain in go.mod.

Done.

reshad-equinix:
> > Please try to use this approach and test it on an internet feed.
> 
> Performance will suffer, by how much I'm not sure. What if I make the synchronous behaviour (to guarantee ordering) enabled via config and the default behaviour stays as-is today?

@sbezverk I have maintained the existing behaviour by default.

sbezverk:
@reshad-equinix  it is possible to use a configuration to flip between sync and async way, but I am still puzzled if what you are trying to fix is not just artificially triggered issue which close to impossible to hit in a real life. Let me check with some BGP developers before final decision.

PR #266, opened 2025-07-15
ensure stream is created on nats service before use

protonjhow:
when attempting to use the NATS Jetstream configuration there were errors in the gobmp logs: 

`1 route-monitor.go:81] failed to process Unicast Prefix message with error: failed to push a message of type 74 to kafka with error: nats: no response from stream`

this appears to be because the stream is not created by the producer. 

here we add the function to create a stream and call it before the end of the publisher setup

Signed-off-by: John Howard <john.howard@proton.ch>

### DIFF
--- a/pkg/nats/nats-publisher.go
+++ b/pkg/nats/nats-publisher.go
@@ -112,6 +112,28 @@ func (p *publisher) Stop() {
 	p.nc.Close()
 }
 
+func (p *publisher) createStreams() error {
+	// Define the stream configuration
+	streamConfig := &nats.StreamConfig{
+		Name:      "goBMP",
+		Subjects:  []string{"gobmp.parsed.*"},
+		Storage:   nats.FileStorage,
+		Retention: nats.InterestPolicy,
+		MaxMsgs:   -1, // No limit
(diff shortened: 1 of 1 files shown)


pkg/nats/nats-publisher.go
```diff
@@ -112,6 +112,28 @@ func (p *publisher) Stop() {
...
+func (p *publisher) createStreams() error {
+	// Define the stream configuration
+	streamConfig := &nats.StreamConfig{
+		Name:      "goBMP",
+		Subjects:  []string{"gobmp.parsed.*"},
+		Storage:   nats.FileStorage,
+		Retention: nats.InterestPolicy,
+		MaxMsgs:   -1, // No limit
+		MaxBytes:  -1, // No limit
+		MaxAge:    24 * time.Hour,
```
sbezverk: currently we have 15 minutes retention time for kafka. 24 hours seems a bit too much imho. Would it be a problem to have the same retention time for nats?

protonjhow: seems totally reasonable to me, and i updated the PR to match

PR #284, opened 2025-12-15
PR-212 BMP RAW + Readme Update

fenilp1:
Implement BMP RAW message support for OpenBMP v2 binary format compatibility and update documentation.

  ## BMP RAW Implementation

  Add support for publishing BMP messages in OpenBMP binary format without parsing BGP content.

  **New Features:**
  - OpenBMP v2 binary header format
  - RAW message type with binary serialization
  - Collector admin ID with MD5 hashing
  - Kafka topic `gobmp.bmp_raw`

  **CLI Flags:**
  - `--bmp-raw={true|false}` - Enable RAW mode (default: false)
  - `--admin-id={string}` - Collector identifier (default: hostname)

  **Testing:**
  - Unit tests for RAW serialization and binary headers
  - Error handling for binary operations
  - Collector hash generation

  ## Documentation Updates

  Updated README with complete parameter reference and deployment guides.

  **Added:**
  - Navigation header
  - Quick Start section with Docker, source build, and Kubernetes options
  - Usage examples
  - CLI parameter reference organized by category
  - Kafka topics table
  - Performance monitoring with pprof
  - Docker Compose example
  - Contributing section

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,243 +1,419 @@
-<p align="left">
+<p align="center">
   <img src="https://github.com/sbezverk/gobmp/blob/master/Hudson_Go_BMP_logo.png?raw=true" width="40%" height="40%">
 </p>
 
+<h1 align="center">goBMP</h1>
 
- 
+<p align="center">
+  <strong>High-performance BGP Monitoring Protocol (BMP) Collector</strong>
+</p>
 
-goBMP is an implementation of Open BMP (RFC 7854) protocol’s collector in go language.  Collected BGP information can be published to kafka topics, stored in the file in JSON format or printed to stdout.
+<p align="center">
+  <a href="#features">Features</a> •
+  <a href="#supported-protocols">Supported Protocols</a> •
+  <a href="#quick-start">Quick Start</a> •
+  <a href="#configuration">Configuration</a> •
+  <a href="#deployment">Deployment</a>
+</p>
 
- 
+---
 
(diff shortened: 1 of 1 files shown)


cmd/gobmp/gobmp.go
```diff
@@ -104,7 +108,25 @@ func main() {
...
+		hostname, err := os.Hostname()
+		if err != nil {
+			glog.Warningf("failed to get hostname, using 'gobmp-collector' as admin ID: %+v", err)
+			collectorAdminID = "gobmp-collector"
+		} else {
+			collectorAdminID = hostname
+		}
+	}
+
+	bmpSrv, err := gobmpsrv.NewBMPServer(srcPort, dstPort, interceptFlag, publisher, splitAFFlag, bmpRawFlag, collectorAdminID)
```
sbezverk, line 129: I do not like the idea of argument list growing, probably soon a gobmp collector config structure should be introduced.

fenilp1: Yes we do need a config.yaml equivalent file.

What are your thoughts on in-memory storage or a hybrid approach with badgerDB?

Today with json dump we don't do any deduplications. The file can keep on growing.

If someone wants to use goBmp as a standalone tool without any external dependencies?

sbezverk: for now just having a config structure would be sufficient,  gobgp does not have too many configuration options like gobgp for example and we do not need to at least at this point to have persistent config in yaml. Let's start with just a structure and then see where it all goes.

PR #297, opened 2025-12-17
Add RFC 9723 (BGP CPR for SRv6) support

fenilp1:
Implement BGP Colored Prefix Routing per RFC 9723.

Changes:
- Add Color field to UnicastPrefix message type
- Extract Color Extended Community from BGP Updates
- Support color-based intent-aware routing for SRv6
- Comprehensive tests for Color EC extraction
- Documentation updates (README, CHANGELOG)

RFC 9723 enables intent-aware routing for SRv6 by associating colors with IPv6 prefixes. Uses existing Color Extended Community parsing (RFC 5512) with minimal changes to IPv6 Unicast handling.

All tests pass. Full RFC 9723 compliance.

### DIFF
--- a/pkg/message/types.go
+++ b/pkg/message/types.go
@@ -84,6 +84,7 @@ type UnicastPrefix struct {
 	IsNexthopIPv4  bool                `json:"is_nexthop_ipv4"`
 	PathID         int32               `json:"path_id,omitempty"`
 	Labels         []uint32            `json:"labels,omitempty"`
+	Color          *uint32             `json:"color,omitempty"` // RFC 9723 BGP Colored Prefix Routing (CPR) for SRv6
 	PrefixSID      *prefixsid.PSid     `json:"prefix_sid,omitempty"`
 	IsEOR          bool                `json:"is_eor,omitempty"`
 	// Values are assigned based on PerPeerHeader flags
@@ -95,6 +96,9 @@ type UnicastPrefix struct {
 }
 
 func (u *UnicastPrefix) Equal(ou *UnicastPrefix) (bool, []string) {
+	if ou == nil {
+		return false, nil
+	}
 	equal := true
 	diffs := make([]string, 0)
 	if u.RouterIP != ou.RouterIP {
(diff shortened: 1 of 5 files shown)


pkg/message/types.go
```diff
@@ -156,6 +157,10 @@ func (u *UnicastPrefix) Equal(ou *UnicastPrefix) (bool, []string) {
 		equal = false
 		diffs = append(diffs, "labels mismatch")
 	}
+	if (u.Color == nil) != (ou.Color == nil) || (u.Color != nil && *u.Color != *ou.Color) {
```
sbezverk: please make this check more human friendly.

fenilp1: Thanks. I'll make the change in the morning.

pkg/message/types.go
```diff
@@ -156,6 +157,14 @@ func (u *UnicastPrefix) Equal(ou *UnicastPrefix) (bool, []string) {
 		equal = false
 		diffs = append(diffs, "labels mismatch")
 	}
+	// Check if Color fields differ: either one is nil and the other isn't, or both are set but different
+	if (u.Color == nil) != (ou.Color == nil) {
+		equal = false
+		diffs = append(diffs, "color mismatch")
+	} else if u.Color != nil && *u.Color != *ou.Color {
```
sbezverk, line 171: would it not generate panic here if u is not nil and ou is nil.  it would be false for check on line 161, then on line 164 when it dereferences *ou.Color it would panic, did I get something wrong? Could you please test this scenarion to confirm.

fenilp1: Line 164 only executes when both pointers have the same nil status (both nil OR both not-nil). The u.Color != nil check combined with Go’s short-circuit evaluation prevents any nil pointer dereference.


Line 164 only runs when line 161 is FALSE. Line 161 is FALSE when:
	∙	BOTH are nil, OR
	∙	BOTH are not nil

Why no panic?
When line 164 checks u.Color != nil && *u.Color != *ou.Color:
	1.	If BOTH are nil → u.Color != nil is false, so it stops (never tries *ou.Color)
	2.	If BOTH are not nil → Both *u.Color and *ou.Color are safe to use

sbezverk: please add check at the beginning of the func for ou == nil then return false, in this case all following dereferences of ou will be safe.

PR #321, opened 2026-01-15
Remove Codecov bot integration

fenilp1:
Remove codecov/codecov-action from CI workflow to reduce external dependencies and PR comment noise. Coverage generation still enabled via `go test -coverprofile=coverage.txt` and uploaded as artifacts.

Coverage can be viewed locally with: go tool cover -html=coverage.txt


fenilp1:
@sbezverk

PR #352, opened 2026-02-19
Implement EVPN Type 8 Multicast Leave Synch route parsing per RFC 9251 Section 9.3

fenilp1:
Implements EVPN Type 8 Multicast Leave Synch route parsing per RFC 9251 Section 9.3

- Paired with Type 7 (Membership Report) for complete join/leave synchronization
- Parse variable-length multicast source, group, and originator addresses
- Parse Reserved and Maximum Response Time fields for IGMP leave timing
- Parse Flags field for leave synchronization control


fenilp1:
@sbezverk can you please start copilot review

PR #358, opened 2026-02-23
Add Type H segment support (IPv6 Local/Remote adjacency)

fenilp1:
Implements SR Policy Segment Type H (type 8) per RFC 9831.

Type H encodes IPv6 local and remote adjacency addresses with an optional MPLS SID. Wire format: 34 bytes (without SID) or 38 bytes (with SID).

**Changes:**
- Add `TypeHSegment` interface and `typeHSegment` struct
- Wire format parsing in `UnmarshalSegmentListSTLV`
- JSON marshal/unmarshal with validation


fenilp1:
@sbezverk ready to merge

PR #364, opened 2026-02-26
fix timestamp decoding issue 363

### LINKED ISSUE #363: Incorrect processing Peer header timestamp
Currently timestamp of any bmp message comes with a timestamp where milliseconds are always 0. It is due to an incorrect parsing of a timestanp in Peer header.

Needs to be fixed to preserve milliseconds value of a rimestamp.

sbezverk:
Closes #363 

PR addresses issue with incorrect decoding millisecond portion of the timestamp and adds a unit test to test it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,13 +28,13 @@ validator-mac:
 	$(MAKE) -C ./cmd/validator compile-validator-mac
 
 container: gobmp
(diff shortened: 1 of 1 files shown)


sbezverk:
@fenilp1  please review this small fix, thank you

PR #426, opened 2026-04-12
Fix SR segment loss, node descriptor bounds, and prefix attr flags panic

fenilp1:
## Summary

- **P3-15 (data loss)**: `UnmarshalSRSegmentListSubTLV` stored every `SRSegment` sub-TLV under the same map key (`SRSegmentType = 1206`), so each segment overwrote the previous one — only the last segment in a Segment List survived. Changed `SRSegmentList.SubTLV` and the function return type from `map[uint16]SRSegmentListSubTLV` to `[]SRSegmentListSubTLV` so all segments are appended in order per RFC 9256 §2.4.4.

- **P3-17 (panic)**: `GetASN`, `GetLSID`, `GetOSPFAreaID`, and `GetConfedMemberASN` in `node-descriptor.go` called `binary.BigEndian.Uint32` without checking `len(tlv.Value) >= 4`, panicking on malformed or truncated TLVs from a misbehaving peer.

- **B4 (panic)**: `UnmarshalPrefixAttrFlags` accessed `b[0]` without a `len(b) == 0` guard, panicking on empty input.

### DIFF
--- a/pkg/base/node-descriptor.go
+++ b/pkg/base/node-descriptor.go
@@ -26,7 +26,7 @@ type NodeDescriptor struct {
 
 // GetASN returns Autonomous System Number used to uniqely identify BGP-LS domain
 func (nd *NodeDescriptor) GetASN() uint32 {
-	if tlv, ok := nd.SubTLV[512]; ok {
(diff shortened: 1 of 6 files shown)


fenilp1:
@sbezverk ready to merge no new comments

sbezverk (approved):

PR #427, opened 2026-04-12
Add passive BMP connection limit

fenilp1:
## Summary

- **SEC-01**: `MaxPassiveConnections` in `config.Config` (YAML: `max_passive_connections`) caps concurrent BMP sessions in passive mode via a channel semaphore. Zero (the default) means no limit. Excess connections are rejected immediately with a Warning log showing the configured limit. Negative values are rejected at `NewBMPServer` with a clear error.

Operator-tunable via YAML and defaults to disabled for backward compatibility. Not applied in active mode (which manages a fixed speaker list).

Note: a per-connection read deadline (SEC-02) is not part of this PR and will be proposed separately.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -67,8 +67,10 @@ type Config struct {
 	SplitAF         *bool        `yaml:"split_af"`
 	BmpListenPort   int          `yaml:"bmp_listen_port"`
 	PerformancePort int          `yaml:"performance_port"` // > 0 enables pprof collection
-	ActiveMode      bool         `yaml:"active_mode"`
-	SpeakersList    []string     `yaml:"speakers_list"`
+	// MaxPassiveConnections caps concurrent BMP sessions in passive mode. 0 means no limit.
+	MaxPassiveConnections int      `yaml:"max_passive_connections"`
+	ActiveMode            bool     `yaml:"active_mode"`
+	SpeakersList          []string `yaml:"speakers_list"`
 }
 
 func LoadConfig(path string) (*Config, error) {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -67,7 +67,8 @@ type Config struct {
 	SplitAF         *bool        `yaml:"split_af"`
 	BmpListenPort   int          `yaml:"bmp_listen_port"`
 	PerformancePort int          `yaml:"performance_port"` // > 0 enables pprof collection
-	// MaxPassiveConnections caps concurrent BMP sessions in passive mode. 0 means no limit.
+	// MaxPassiveConnections caps concurrent passive-mode BMP TCP sessions (0 = unlimited).
+	// YAML key is max_passive_connections (not max_connections). Ignored in active_mode.
 	MaxPassiveConnections int      `yaml:"max_passive_connections"`
 	ActiveMode            bool     `yaml:"active_mode"`
 	SpeakersList          []string `yaml:"speakers_list"`
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -78,8 +78,10 @@ type Config struct {
 	SplitAF         *bool        `yaml:"split_af"`
 	BmpListenPort   int          `yaml:"bmp_listen_port"`
 	PerformancePort int          `yaml:"performance_port"` // > 0 enables pprof collection
-	ActiveMode      bool         `yaml:"active_mode"`
-	SpeakersList    []string     `yaml:"speakers_list"`
+	// MaxPassiveConnections caps concurrent BMP sessions in passive mode. 0 means no limit.
+	MaxPassiveConnections int      `yaml:"max_passive_connections"`
+	ActiveMode            bool     `yaml:"active_mode"`
+	SpeakersList          []string `yaml:"speakers_list"`
 }
 
 func LoadConfig(path string) (*Config, error) {
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -78,7 +78,8 @@ type Config struct {
 	SplitAF         *bool        `yaml:"split_af"`
 	BmpListenPort   int          `yaml:"bmp_listen_port"`
 	PerformancePort int          `yaml:"performance_port"` // > 0 enables pprof collection
-	// MaxPassiveConnections caps concurrent BMP sessions in passive mode. 0 means no limit.
+	// MaxPassiveConnections caps concurrent passive-mode BMP TCP sessions (0 = unlimited).
+	// YAML key is max_passive_connections (not max_connections). Ignored in active_mode.
 	MaxPassiveConnections int      `yaml:"max_passive_connections"`
 	ActiveMode            bool     `yaml:"active_mode"`
 	SpeakersList          []string `yaml:"speakers_list"`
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -67,8 +67,12 @@ type Config struct {
 	SplitAF         *bool        `yaml:"split_af"`
 	BmpListenPort   int          `yaml:"bmp_listen_port"`
 	PerformancePort int          `yaml:"performance_port"` // > 0 enables pprof collection
-	ActiveMode      bool         `yaml:"active_mode"`
-	SpeakersList    []string     `yaml:"speakers_list"`
+	// MaxConnections caps concurrent BMP sessions in passive mode. 0 means no limit.
+	MaxConnections int `yaml:"max_connections"`
+	// ReadTimeoutSec sets a per-read deadline (seconds) on BMP connections. 0 means no timeout.
```
sbezverk: I think if used it will cause harm. bgp is relatively stable protocol, in some cases bgp might not be sending updates for days or even more.  I do not see value in using this time out, unless I misunderstood its purpose.

pkg/config/config.go
```diff
@@ -67,8 +67,12 @@ type Config struct {
 	SplitAF         *bool        `yaml:"split_af"`
 	BmpListenPort   int          `yaml:"bmp_listen_port"`
 	PerformancePort int          `yaml:"performance_port"` // > 0 enables pprof collection
-	ActiveMode      bool         `yaml:"active_mode"`
-	SpeakersList    []string     `yaml:"speakers_list"`
+	// MaxConnections caps concurrent BMP sessions in passive mode. 0 means no limit.
+	MaxConnections int `yaml:"max_connections"`
```
sbezverk: if it is just for passive mode, probably make sense to change name of this var so it would be more intuitive.

sbezverk (changes requested):

pkg/config/config.go
fenilp1: You are right — removed entirely in 9006690. BMP has no application-level keepalives, so a stable BGP session with no route changes produces zero BMP traffic. The rolling deadline would fire on any quiet period, disconnecting a healthy session.

pkg/config/config.go
fenilp1: Renamed to MaxPassiveConnections (yaml: max_passive_connections) in 9006690.

pkg/gobmpsrv/gobmpsrv.go
```diff
@@ -21,15 +21,6 @@ import (
 // maxBMPMessagePayload is the maximum allowed BMP message payload size (1 MB).
 const maxBMPMessagePayload = 1 << 20
 
-// stableSessionThreshold is the minimum duration the bmpWorker session must run
```
sbezverk, line 24: why it is removed? `must be restired as it is addressing encountered issue when connecting through proxy.

fenilp1: Good catch. The refactor accidentally removed stableSessionThreshold and reset retryDelay to 1 s on every TCP connect, which defeated the proxy protection entirely. Restored in b4be9c4 — your original logic is back: session duration >= 30 s resets retryDelay to 1 s, anything shorter doubles it (proxy scenario). Also added tests for both paths.

fenilp1:
@sbezverk review please

pkg/gobmpsrv/gobmpsrv.go
```diff
@@ -146,6 +159,10 @@ func (srv *bmpServer) startWorker(client net.Conn) {
 	if srv.closing {
 		srv.mu.Unlock()
 		_ = client.Close()
+		// Release semaphore slot acquired by server() before this call.
+		if srv.connSem != nil {
```
sbezverk: I think this semaphore release is too implicit. startWorker() is releasing connSem based only on connSem != nil, but it has no proof that this particular call actually acquired a slot. Today that is true only because server() acquires before calling startWorker(), and the tests manually pre-fill the channel the same way.

That makes the code fragile: any future direct call to startWorker() with connSem != nil but without a prior acquire can block forever on <-srv.connSem> or drain another connection's token.

I’d suggest making ownership explicit, e.g. pass a hasSem flag into startWorker(), or move acquire/release into the same function so the invariant is local and enforced.

sbezverk (changes requested):

pkg/gobmpsrv/gobmpsrv.go
fenilp1: Addressed in 32c64cb: startWorker now takes a hasSem bool; only the server() call that acquired the slot passes true, so release is tied to actual ownership.

pkg/gobmpsrv/gobmpsrv_test.go
```diff
@@ -1348,75 +1323,248 @@ func TestConnectSpeaker_ShortLivedSession_BackoffCapped(t *testing.T) {
...
+
+	srv.Start()
+	defer func() {
+		// Drain the semaphore so Stop() does not hang waiting for bmpWorker exits.
+		select {
+		case <-srv.connSem:
+		default:
+		}
+		srv.Stop()
+	}()
```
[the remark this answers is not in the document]
fenilp1: Addressed in 3c184ec: replaced the drain+comment with a plain `defer srv.Stop()` — the drain was dead code (semaphore was pre-filled only to force the reject path; Stop() never reads from connSem).

fenilp1:
@sbezverk review please

fenilp1:
@sbezverk review please

sbezverk:
@fenilp1 
Thanks for narrowing this PR to the passive BMP connection cap and removing the read-timeout part. The current implementation is much cleaner than the earlier version, defaults to disabled, and the semaphore ownership is now explicit.
That said, I still see this as an optional hardening knob rather than a core BMP collector feature. It only bounds passive-mode resource usage; it does not by itself prevent one noisy source from occupying all available slots. If we merge it, I think we should make the operator-facing behavior fully documented and config-tested first.
Comment 1: README / Operator Docs
max_passive_connections is a new YAML-facing operator knob, but it is not listed in the README “All available fields” config reference.
Since this option has no CLI flag, README/config docs are the main discoverability path. Please add it near the passive/active-mode fields, with:
default 0
0 means unlimited
passive-mode only; ignored in active mode
behavior when the limit is reached: new TCP session is rejected/closed immediately
Comment 2: Config Test Coverage
Please add focused LoadConfig coverage for max_passive_connections.
The gobmpsrv tests cover constructed config.Config values, but I do not see a test proving the actual YAML key unmarshals into Config.MaxPassiveConnections. I’d suggest one positive YAML test and one omitted/default case verifying it remains 0.
Comment 3: Semaphore Release Ordering
Small lifecycle nit in startWorker: the defer currently calls srv.wg.Done() before releasing the connSem slot.
That is probably harmless in practice, but it means Stop() can technically stop waiting before the limiter slot has been released. I’d prefer releasing connSem before wg.Done() so “worker is done” also means its connection-limit accounting is complete.

thank you

fenilp1:
@sbezverk review please

sbezverk (approved):

PR #447, opened 2026-05-20
Preserve unrecognised BGP path attributes (RFC 4271 §5)

fenilp1:
Unknown attribute Types (43-127, 129-255 excluding handled cases) were silently dropped from BaseAttributes. Adds default branch in unmarshalBaseAttrsFromSlice capturing Type, full Flags byte, and raw Value into UnknownAttributes (json:omitempty), plus comparison in BaseAttributes.Equal so updates differing only in unknown attrs aren't deduped silently.


fenilp1:
@sbezverk review please

fenilp1:
@sbezverk review please

sbezverk:
@fenilp1 Unfortunately I ran out of Co-Pilot credits for June, so will have to review it when I have cycles. Sorry for the delay.

fenilp1:
@sbezverk review please

sbezverk (approved):

PR #450, opened 2026-06-01
Add --kafka-skip-topic-creation flag for Kafka 4.0+ compatibility

fenilp1:
Kafka 4.0+ brokers reject the Admin API \`CreateTopics\` call with \`UnsupportedVersionException\`. Many production clusters (Confluent Cloud, MSK, hardened on-prem) also disable topic auto-creation, making gobmp fail on startup with no workaround.

Add \`--kafka-skip-topic-creation=true\` to bypass the \`ClusterAdmin\` init and \`ensureTopic\` loop entirely. When set, gobmp connects directly as an async producer; topics must be pre-created before starting gobmp.

Also fixes a pre-existing resource leak: the controller broker connection opened during startup was logged but never closed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,23 @@ Examples:
 Kafka topic retention time in milliseconds. Topics are created with this retention policy to manage storage for high-volume BGP data. Adjust based on your storage capacity and retention requirements.
 
 ```
+--kafka-skip-topic-creation={true|false}
+```
+**Default:** false
+
+When `true`, gobmp skips the Kafka Admin API `CreateTopics` call on startup and connects directly as an async producer. Use this with Kafka 4.0+ environments that reject `CreateTopics`, or with clusters where the client lacks `CreateTopics` permission (e.g. Confluent Cloud, MSK with restricted IAM policies).
+
+> **Important:** Topics must be pre-created before starting gobmp. If the required topics do not exist, publishing will fail with unclear broker errors.
+
+YAML equivalent:
+```yaml
+kafka_config:
+  kafka_srv: "kafka.example.com:9092"
+  kafka_tp_retn_time_ms: 900000
+  skip_topic_creation: true
+```
+
+```
 --nats-server={url}
 ```
 **Default:** none (NATS disabled)
--- a/cmd/gobmp/gobmp.go
+++ b/cmd/gobmp/gobmp.go
@@ -22,18 +22,19 @@ import (
 )
 
 var (
-	srcPort           int
-	perfPort          int
-	kafkaSrv          string
-	kafkaTpRetnTimeMs string // Kafka topic retention time in ms
-	kafkaTopicPrefix  string
-	natsSrv           string
-	splitAF           string
-	dump              string
-	file              string
-	bmpRaw            string
-	adminID           string
-	configFile        string
+	srcPort                int
+	perfPort               int
+	kafkaSrv               string
+	kafkaTpRetnTimeMs      string // Kafka topic retention time in ms
+	kafkaTopicPrefix       string
+	kafkaSkipTopicCreation string
+	natsSrv                string
+	splitAF                string
+	dump                   string
+	file                   string
+	bmpRaw                 string
+	adminID                string
+	configFile             string
 )
 
 const (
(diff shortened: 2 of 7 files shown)


fenilp1:
@sbezverk review please

sbezverk (changes requested):
## 🔴 Critical Issues

**1. Variable shadowing and potential nil dereference**
- Line 217 in kafka-publisher.go: `var producer sarama.AsyncProducer` declares an interface type that will be nil if not assigned. While you do assign it immediately on line 218, this pattern is risky and makes the code harder to verify. Consider initializing inside the conditional or use a different pattern.

**2. Inconsistent resource cleanup on error paths**
- In SkipTopicCreation=true path (else branch line 214): if NewAsyncProducer fails, neither ca nor cb get closed (both nil), but this asymmetry with the other path makes it confusing.
- In SkipTopicCreation=false path: if ensureTopic fails at line 209, you close ca at line 210, but cb was already closed at line 205. The flow is hard to follow and verify correctness.

**3. Missing validation that topics exist when skipping creation**
- When SkipTopicCreation=true, there is no validation that the required topics actually exist before the publisher starts. If topics do not exist, publishing will fail later with unclear error messages. Consider adding pre-flight validation or at least better documentation warning users.

## 🟡 Moderate Issues

**4. Test-scoped package variables cause potential race conditions**
- Lines 51-57 in kafka-publisher.go expose adminRetryMax and netReadTimeout as package-level variables for tests to modify. If tests run in parallel, these can cause race conditions. Consider using dependency injection or passing through Config struct instead.

**5. Code complexity and clarity in NewKafkaPublisher**
- Function is now 49 lines with heavily nested conditionals, making error handling paths hard to verify:
  - ca is declared but only conditionally assigned
  - cb is only used in one path then closed
  - multiple resource cleanup points increase leak risk
- Consider extracting the topic-creation logic into a separate function (for example ensureTopics()).

**6. Incomplete error handling**
- Line 201-202: if waitForControllerBrokerConnection fails, you close ca but do not provide context about the broker connection failure in logs.
- Missing comprehensive validation beyond strconv.ParseBool check for the flag.

## 🟡 Minor Issues

**7. Uncovered code paths (per Codecov)**
- 81.57895% coverage on kafka-publisher.go (4 missing lines, 2 partials)
- 88.88% on gobmp.go (1 missing line)
- Likely uncovered paths: some SkipTopicCreation=false or error-path branches.

**8. Documentation gaps**
- NewKafkaPublisher function comment does not explain SkipTopicCreation=true behavior.
- No warning that topics must be pre-created externally or publishing may fail later.
- Flag help text could be clearer about who must pre-create topics and when validation occurs.

**9. Error handling style inconsistency**
- Line 205: `_ = cb.Close()` silently ignores close errors.
- Line 210: `_ = ca.Close()` same issue.
- Consider logging close failures with warning level.

**10. Control flow clarity**
- Lines 214-215 use an else branch with a log message; readability would improve by checking SkipTopicCreation in a positive guard first and keeping topic-creation path in the else.

## ✅ What is Good
- Proper nil-check in Stop method (line 161-163).
- Good test coverage for new flag parsing.
- Clear commit progression.
- Tests cover both skip and non-skip paths.

## 📋 Recommendations

1. **High priority**: Extract topic-creation logic into a helper function to reduce NewKafkaPublisher complexity.
2. **High priority**: Add a validation function that checks required topics exist when SkipTopicCreation=true.
3. **High priority**: Remove package-level mutable test knobs; prefer dependency injection or config-driven values.
4. **Medium priority**: Add an integration test verifying behavior when required topics are missing.
5. **Medium priority**: Improve error messages to distinguish missing-topic failures from other broker/producer failures.

fenilp1:
@sbezverk review please.

sbezverk (changes requested):
Please add README documentation for --kafka-skip-topic-creation and YAML kafka_config.skip_topic_creation, including an example and note that topics must be pre-created. 

Also please soften “Required for Kafka 4.0+” to avoid implying all Kafka 4.0 brokers reject CreateTopics.  Something along these lines: this to “use with Kafka 4.0+ environments that reject CreateTopics” or “clusters where CreateTopics is unsupported/disabled.”

Finally, please add a Kafka mock test where SkipTopicCreation=false handles CreateTopicsRequest successfully and returns a working publisher, so the normal topic-creation startup path is covered.

sbezverk (approved):

PR #465, opened 2026-07-27
kafka: support TLS connections to the broker

tat3rb:
## Problem

`NewKafkaPublisher` builds its sarama config with `Net.TLS` left at the zero value, and `kafka.Config` has no field to change that. gobmp can therefore only reach a plaintext broker. A broker that terminates TLS on a dedicated listener refuses plaintext outright, so it is unreachable today with no configuration surface to fix it.

## Change

Adds one-way (server-authenticated) TLS:

```
--kafka-tls        enable TLS on broker connections
--kafka-ca <path>  PEM bundle used to verify the broker certificate;
                   empty falls back to the system trust store
```

Both are also settable from the YAML config as `kafka_tls` / `kafka_ca`, matching how the other Kafka options are plumbed.

The client presents no certificate, which fits brokers that use network scope rather than client certificates for authorization. `MinVersion` is TLS 1.2.

**Defaults are unchanged** — `--kafka-tls` is off, so existing plaintext deployments behave exactly as before.

## Notes

`newTLSConfig` checks the boolean returned by `x509.CertPool.AppendCertsFromPEM`. It reports failure with a bool rather than an error and silently ignores unparsable blocks, so a truncated or mistaken bundle would otherwise produce an empty pool and surface much later as an opaque "unknown authority" rather than naming the file that was wrong.

## Tests

`pkg/kafka/tls_test.go` is table-driven over the failure class rather than just the happy path: valid CA, two-certificate chain, certificate with a text preamble (openssl emits this and it must not be treated as corruption), nonexistent path, empty file, non-PEM garbage, PEM containing no certificate (pointing at the key by mistake), truncated certificate, and a directory instead of a file. It also asserts the client never presents a certificate and that `InsecureSkipVerify` is never set.

`go build ./...` and `go test ./...` pass.

Happy to adjust naming or split this differently if you'd prefer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -157,6 +157,22 @@ kubectl get pods -l app=gobmp
   --split-af=true
 ```
 
+### Kafka Publishing over TLS
+Brokers that terminate TLS on a dedicated listener (commonly `:9094`) refuse
+plaintext connections, so `--kafka-tls` is required to reach them:
+```bash
+./bin/gobmp --source-port=5000 \
+  --kafka-server=kafka.example.com:9094 \
+  --kafka-tls \
+  --kafka-ca=/etc/pki/tls/certs/example-root-ca.pem \
+  --split-af=true
+```
+`--kafka-ca` is optional. Omit it when the broker certificate chains to a CA
+already in the host trust store:
+```bash
+./bin/gobmp --source-port=5000 --kafka-server=kafka.example.com:9094 --kafka-tls
+```
+
 ### NATS Publishing
 ```bash
 ./bin/gobmp --source-port=5000 \
@@ -203,6 +219,16 @@ kafka_config:
   admin_id: "collector-01"
 ```
 
+**Kafka over TLS (`config.yaml`):**
+```yaml
+bmp_listen_port: 5000
+kafka_config:
+  kafka_srv: "kafka.example.com:9094"
+  kafka_tls: true
+  # Optional. Omit to verify against the host trust store.
+  kafka_ca: "/etc/pki/tls/certs/example-root-ca.pem"
+```
+
 **NATS example:**
 ```yaml
 bmp_listen_port: 5000
@@ -227,6 +253,8 @@ kafka_config:
   kafka_srv: "host:port"     # required to activate Kafka publisher
   kafka_tp_retn_time_ms: 900000
   kafka_topic_prefix: ""     # optional topic name prefix
+  kafka_tls: false           # true = connect to the broker over TLS
+  kafka_ca: ""               # PEM CA bundle; empty = host trust store
   bmp_raw: false             # OpenBMP RAW mode
   admin_id: ""               # defaults to OS hostname
 
@@ -320,6 +348,33 @@ Full path and filename for storing processed messages when `--dump=file` is used
 **Default:** none (Kafka disabled)
 
 Kafka broker address for publishing BMP messages. When specified, goBMP publishes parsed messages to topic-specific Kafka topics. Example: `--kafka-server=kafka.example.com:9092`
+
+```
+--kafka-tls
+```
+**Default:** false (plaintext)
+
+Connect to the Kafka broker over TLS. Required for brokers that expose a
+TLS-only listener, conventionally `:9094`, since those refuse plaintext
+connections outright.
+
+This is one-way (server-authenticated) TLS: goBMP verifies the broker
+certificate and presents none of its own, which suits brokers that use network
+scope rather than client certificates for authorization. The negotiated
+protocol floor is TLS 1.2.
+
+```
+--kafka-ca={path}
+```
+**Default:** empty (host trust store)
+
+Path to a PEM bundle used to verify the broker certificate. Only consulted when
+`--kafka-tls` is set.
+
+Leave it empty when the broker certificate chains to a CA the host already
+trusts; supply it when the broker uses a private CA that is not installed as a
+system trust anchor. An unreadable file, or one containing no certificate, is
+rejected at startup before any connection is attempted.
 
 ```
 --kafka-topic-prefix={prefix}
--- a/cmd/gobmp/gobmp.go
+++ b/cmd/gobmp/gobmp.go
@@ -28,6 +28,8 @@ var (
 	kafkaTpRetnTimeMs      string // Kafka topic retention time in ms
 	kafkaTopicPrefix       string
 	kafkaSkipTopicCreation string
+	kafkaTLS               bool
+	kafkaCA                string
 	natsSrv                string
 	splitAF                string
 	dump                   string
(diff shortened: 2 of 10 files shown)


sbezverk:
@tat3rb 
Thank you for adding Kafka TLS support. I reviewed the code path and the TLS helper looks semantically correct to me: TLS defaults remain off, `newTLSConfig` uses TLS 1.2 minimum, does not set `InsecureSkipVerify`, does not present client certificates, and correctly validates custom CA PEM input.

A few small asks before merge:

1. Please document the new options in `README.md`.

   The README Kafka examples and command-line reference should mention:

   - CLI: `--kafka-tls`, `--kafka-ca`
   - YAML: `kafka_config.kafka_tls`, `kafka_config.kafka_ca`

   Please include a short example, especially noting that empty `kafka_ca` uses the system trust store.

2. Please add focused tests for config plumbing.

   `cmd/gobmp/gobmp.go` adds override handling for `kafka-tls` and `kafka-ca`, but `cmd/gobmp/gobmp_test.go` does not currently register those flags in `newTestFlagSet` or test the override behavior. Please add tests proving CLI flags populate `KafkaConfig.KafkaTLS` and `KafkaConfig.KafkaCA`.

   It would also be useful to add a YAML load test proving `kafka_tls` and `kafka_ca` unmarshal into `pkg/config.KafkaConfig`.

3. Please add at least one publisher-path test.

   `pkg/kafka/tls_test.go` covers `newTLSConfig` well, but `NewKafkaPublisher`’s TLS branch is not exercised. A small test using `TLS: true` with an invalid CA path would prove the publisher fails early with the intended CA-file error before attempting Kafka connections. A TLS mock broker success path would be even better, but the invalid-CA test is probably enough as focused coverage.

I ran:

- `go test -count=1 ./pkg/kafka ./cmd/gobmp ./pkg/config`
- `go test -count=1 ./...`
- `go build ./...`

All passed locally when tests were run with loopback binding allowed.

sbezverk (changes requested):
See comments added to PR discussion.

tat3rb:
Thanks for the quick and careful review. All three addressed in `40ed7a9`.

**1. README.** Documented `--kafka-tls` and `--kafka-ca` in the command-line reference, added a TLS usage example alongside the existing plaintext one, added a YAML example, and added both keys to the annotated `config.yaml` reference. Each place notes that an empty `kafka_ca` verifies against the host trust store, with a one-liner example showing the flag omitted entirely.

**2. Config plumbing.** You were right that `newTestFlagSet` never registered the flags, so the override handling was untested. Both are registered now, with tests proving the CLI flags populate `KafkaConfig.KafkaTLS` and `KafkaConfig.KafkaCA`.

I added two cases beyond the ask, since they pin properties that would be easy to regress:

- an invocation that never mentions either flag leaves TLS off, so existing plaintext deployments are unaffected;
- an explicit `--kafka-tls=false` overrides a config file that enables it, so an operator can turn TLS off without editing the file.

Plus the YAML load tests in `pkg/config`, including one where both keys are absent and the config still loads with TLS off.

**3. Publisher path.** Added `TestNewKafkaPublisherTLSBadCA`, covering three unusable CA inputs: a nonexistent path, a file containing no certificate, and an empty file.

One note on how it is written. The server address is `192.0.2.1:9094` — TEST-NET-1, RFC 5737, guaranteed unroutable — and the test asserts on elapsed time as well as on the error. That assertion is the substance of it: if the CA were validated after the connection attempt rather than before, the call would block for `brockerConnectTimeout` plus metadata retries instead of returning immediately. All three cases return in ~0s, which is the property you were asking about. Happy to drop the timing assertion if you would rather keep the test to a plain error check.

`newTLSConfig` is now at 100% statement coverage and `applyConfigOverrides` at 89.2%. `NewKafkaPublisher` is at 33.3% — the TLS-rejection branch is covered, but the success path still needs a broker, so a mock-broker test is the remaining gap there if you want it.

Ran your commands, all green:

- `go test -count=1 ./pkg/kafka ./cmd/gobmp ./pkg/config`
- `go test -count=1 ./...`
- `go build ./...`

sbezverk:
@tat3rb you need to rebase your PR also try to hit at least 70% coverage.

tat3rb:
@sbezverk Thanks, both done: rebased, and I went with the mock-broker option you suggested.

**Rebase.** Now on top of `bfce930`. The overlap was with #450, which landed in the same Kafka plumbing: `Config`, `KafkaConfig`, the flag block, the `applyConfigOverrides` switch, and the `kafka.Config` literal in `main()`. Every conflict was two additions in one place, so they all resolved additively and nothing from `--kafka-skip-topic-creation` changed behavior. The TLS block in `NewKafkaPublisher` sits ahead of the new cluster-admin path, so the two features are independent.

**Publisher path.** I took the TLS mock broker rather than stopping at the invalid-CA check, since you mentioned it would be better. `sarama.NewMockBroker` is plaintext only, so the tests front it with a TLS-terminating proxy using an in-test self-signed cert, with the CA file as the trust anchor. One detail that might be useful if this pattern gets reused: the mock has to advertise the proxy address in its `MetadataResponse`, because sarama dials the advertised address for the producer's own connection. Advertising `mb.Addr()` sends the producer around the TLS listener and the test passes without a handshake ever happening. The proxy counts completed handshakes and the tests assert on that, so that particular false pass is closed off.

There are two success cases, because they cover different connections:

- `TestNewKafkaPublisherTLSMockBroker`, `SkipTopicCreation=true`, covers the producer connection.
- `TestNewKafkaPublisherTLSTopicCreation`, `SkipTopicCreation=false`, which is what a TLS deployment gets by default. `ClusterAdmin` and then the controller-broker connection are both opened from the same sarama config the TLS branch populated, so both come up over TLS before any producer exists. It asserts more than one handshake so it cannot pass on the producer alone.

Both bound the call in a `select` instead of calling it directly. If the TLS branch ever regresses, the client speaks Kafka at a TLS listener, the handshake never completes, and sarama retries metadata 300 times on a 10s backoff, so a plain call would hang until the package timeout and surface as an unrelated panic. I confirmed the behaviour by mutating `config.Net.TLS.Enable` to `false`: the test fails in 20s and names the cause.

I also added `TestNewTLSConfigVerifiesPeer`, which dials a handshake-only TLS server with a matching CA and with an unrelated one. The negative case is the point: it pins that verification is genuinely enforced, so it breaks if `InsecureSkipVerify` ever appears or an empty pool is allowed to stand in for a trust store.

**Coverage.** Measured against `bfce930` using the CI invocation (`go test -race -coverprofile=coverage.txt -covermode=atomic ./...`):

- patch coverage of added lines: **95.6%** (43/45)
- `newTLSConfig`: 100%
- `NewKafkaPublisher`: 80.8% to 83.3%
- `applyConfigOverrides`: 89.8%
- project total: 76.4881% to 76.5489%, so it moves up slightly

The only added lines still uncovered are the two `kafka.Config` fields in `main()`, in the same literal as the existing `SkipTopicCreation` field and not reachable without running `main`. Glad to add a seam there if you want it, though that would be a change to the existing structure rather than to this feature.

One note in case it saves you a minute: the Codecov comment above is still scored against `038a69c`, the first commit, which is where the 45.45% comes from. Later pushes have not had a `cicd` run to score, since fork PRs wait on maintainer approval, so approving the run on the current head should refresh it. Locally the numbers are the ones above.

Commands, all green:

- `go build ./...`
- `go test -count=1 ./pkg/kafka ./cmd/gobmp ./pkg/config`
- `go test -count=1 ./...`
- `go test -race -coverprofile=coverage.txt -covermode=atomic ./...`
- `golangci-lint run` (v2.8.0, 0 issues) and `go mod tidy` clean

sbezverk (approved):

PR #468, opened 2026-08-09
message: take speaker identity from the connection when Peer Up has none

Vadims06:
The producer latches the BMP speaker's identity on the first Peer Up it sees,
taking it from that message's Local Address. RFC 9069 Section 5.2 zero-fills
that field for a Loc-RIB Instance Peer, so a speaker whose first Peer Up
describes the Loc-RIB latches on `0.0.0.0`, and every message it produces then
carries `router_ip "0.0.0.0"` and `router_hash md5("0.0.0.0")`.

Two such speakers are indistinguishable. A collector keyed on `router_hash`
merges their tables: observations of the same peer collapse onto a single
entry, and a Peer Down from one speaker flushes routes the other is still
reporting. Measured against a lab with two route reflectors, a 20-route
baseline arrived as 12 routes and produced 8 spurious change events; with this
fix the same run reports 20 routes across 2 distinct sources and no spurious
events.

Fall back to the remote address of the TCP connection, which the server
already resolves per client and passes down as `bmp.Message.SpeakerIP` for RAW
messages. A Peer Up that does carry a local address still wins, so output is
unchanged for every session that was already identified correctly.

The latch itself always completes. Route monitor and stats goroutines block on
`speakerReady` before producing anything, so declining to latch on an
uninformative Peer Up would stall them for the life of the connection; when
neither the Peer Up nor the connection names a host, the Peer Up value is kept
rather than an identity invented.

## Test plan
- [x] `go build ./...`
- [x] `go test ./pkg/message/...` — new `speaker_identity_test.go` covers: local
      address wins, IPv4/IPv6 zero-fill falls back to the connection address,
      two Loc-RIB-only speakers stay distinct, established sessions are
      unchanged, and the case where neither address is available still lets
      the producer latch (no goroutine block).

### DIFF
--- a/pkg/message/peer.go
+++ b/pkg/message/peer.go
@@ -62,7 +62,7 @@ func (p *producer) producePeerMessage(op int, msg bmp.Message) {
 		// first PeerUp the fields are immutable — no data race with readers
 		// that wait on speakerReady.
 		p.speakerReadyOnce.Do(func() {
-			p.speakerIP = m.LocalIP
+			p.speakerIP = speakerAddress(m.LocalIP, msg.SpeakerIP)
 			md5Sum := md5.Sum([]byte(p.speakerIP))
 			p.speakerHash = hex.EncodeToString(md5Sum[:])
 			close(p.speakerReady)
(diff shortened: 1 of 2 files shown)


sbezverk (changes requested):
Please confirm references, other than that looks good.

Vadims06:
Thanks. I rechecked the references and corrected the PR description, code comment, and test comment. The previous section number was wrong: RFC 9069 Section 4.2 defines the Loc-RIB peer flags. The zero-filled Local Address in a Loc-RIB Peer Up is specified in [RFC 9069 Section 5.2](https://www.rfc-editor.org/rfc/rfc9069.html#section-5.2), which clarifies [RFC 7854 Section 4.10](https://www.rfc-editor.org/rfc/rfc7854.html#section-4.10). [RFC 9069 Section 6.1.1](https://www.rfc-editor.org/rfc/rfc9069.html#section-6.1.1) also says that a Loc-RIB instance is identified by the Peer Distinguisher and BGP ID.

I also noticed your per-session guidance in #470. If `router_ip` must come from each session's Peer Up, a Type 3 Loc-RIB session supplies a zero-filled Local Address as required by Section 5.2. For Type 3, should `router_ip` fall back to the BMP TCP connection address, or should the identity be derived from the BGP ID and Peer Distinguisher?

sbezverk:
For Type 3, router_ip will be assigned value of SpeakerIP, and SpeakerIP is derived from TCP Connection when PeerUP message is processed.

Vadims06:
Thanks for confirming. This matches the current PR behavior: for Type 3, the zero-filled Peer Up Local Address falls back to `msg.SpeakerIP`, which is derived from the TCP connection. The RFC references are now corrected. Please re-review when convenient.

sbezverk (approved):

PR #470, opened 2026-08-13
Change peer tracking speakerIP assignment from m.LocalIP to msg.SpeakerIP

ZacharyPuls:
After running gobmp against a production router with ~400 or so peers, I noticed that `router_ip` and `router_hash` were being populated with a random IP address/hash, which appeared to be just the first up peer it processed.

This PR updates `pkg/message/peer.go` to change `p.speakerIP` from `m.LocalIP` to `msg.SpeakerIP`, given that  `p.speakerIP` should be populated from the BMP speaker's IP rather than the first connected peer's `LocalIP`. This fixes `router_ip` from being populated incorrectly.

### DIFF
--- a/pkg/message/peer.go
+++ b/pkg/message/peer.go
@@ -62,7 +62,7 @@ func (p *producer) producePeerMessage(op int, msg bmp.Message) {
 		// first PeerUp the fields are immutable — no data race with readers
 		// that wait on speakerReady.
 		p.speakerReadyOnce.Do(func() {
-			p.speakerIP = m.LocalIP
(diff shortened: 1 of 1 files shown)


sbezverk:
@ZacharyPuls  you have correctly identified the problem, unfortunately the proposed solution is not correct. The correct solution needs to keep a table of sessions, and all messages for a specific session needs to use router_ip from session's PeerUp message. If you want to take a stab at it, please do otherwise I will fix it in a few days. Please advise if you decide to work on it.

sbezverk:
@ZacharyPuls  closing it as it has been addressed by #475

PR #473, opened 2026-08-31
draft-ietf-idr-nhc

SergeKrier:
Previously, path attribute type 39 had an empty switch case. NHC data was parsed as a raw path attribute but then silently omitted from  BaseAtt
 ributes  and published JSON.

 Changes

   • Added NHC attribute decoding:
     • AFI, SAFI, and next-hop address
     • 16-bit characteristic code and length
     • Unknown and out-of-order characteristics
     • Extended-length path attributes
   • Added BGPID characteristic support:
     • Validates 8-byte length and nonzero identifier
     • Decodes BGP identifier and four-byte ASN
     • Discards malformed and duplicate instances
   • Added structural validation and RFC 7606-style attribute discard.
   • Added semantic next-hop matching:
     • IPv4 and IPv6
     • RFC 8950 IPv6 next hops for IPv4 NLRI
     • IPv6 global plus link-local forms
     • VPN/RD-prefixed next hops
     • Link-local-only BGPID/ASN validation
   • Route-monitor processing now discards stale NHC attributes when next hops mismatch.
   • Adj-RIB-Out/Loc-RIB link-local attributes are retained with  nhc_unverified  when authoritative sender identity is unavailable.
   • Raw NHC bytes remain available through  nhc_attr  for forensic use.
   • Added NHC handling to  BaseAttributes.Equal .
   • Prefer MP_REACH over MP_UNREACH when both occur, preventing withdrawals from incorrectly driving NHC validation.

 Implementation: bgp-nhc.go

 Base attribute integration: bgp-base-attributes.go:427-435

 Route-context validation: route-monitor.go:109-133

 Tests: bgp-nhc_test.go nhc_test.go

 Verification

   •  go test ./...  — passed
   •  go vet ./...  — passed
   •  git diff --check  — passed

### DIFF
--- a/pkg/bgp/bgp-base-attributes.go
+++ b/pkg/bgp/bgp-base-attributes.go
@@ -71,6 +71,11 @@ type BaseAttributes struct {
 	LgCommunityList []string      `json:"large_community_list,omitempty"`
 	BGPPrefixSID    *BGPPrefixSID `json:"bgp_prefix_sid,omitempty"`
 	OTC             uint32        `json:"otc,omitempty"` // RFC 9234 Only to Customer (OTC) Attribute (Type 35)
+	NHCAttr         []byte        `json:"nhc_attr,omitempty"`
+	NHC             *NHC          `json:"nhc,omitempty"`
+	NHCMalformed    bool          `json:"nhc_malformed,omitempty"`
+	NHCDiscarded    bool          `json:"nhc_discarded,omitempty"`
+	NHCUnverified   bool          `json:"nhc_unverified,omitempty"`
 	// SecPath
 	AttrSet *AttrSet `json:"attr_set,omitempty"` // RFC 6368 ATTR_SET Attribute (Type 128)
 	// UnknownAttributes preserves any path attribute whose Type code is not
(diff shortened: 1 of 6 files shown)


sbezverk (changes requested):
1. `GetNLRIType()` now prefers `MP_REACH_NLRI` over `MP_UNREACH_NLRI`. Since `produceRouteMonitorMessage()` processes only the single attribute returned by `GetNLRIType()`, an UPDATE carrying both reachable and unreachable MP NLRI would now publish only the reachable NLRI and silently skip the withdrawal side. Please split the concerns: NHC validation can prefer MP_REACH for next-hop validation, but route publication should not lose MP_UNREACH data. A focused test with both MP_REACH and MP_UNREACH in one UPDATE would be helpful.

2. `UnmarshalNHC()` currently accepts and publishes an NHC with zero characteristic TLVs, and the test locks that behavior in. `draft-ietf-idr-nhc-07` says an NHC with no characteristics may be considered malformed and must not be further propagated. For gobmp, preserving `nhc_attr` for forensics is fine, but I don’t think we should emit a decoded `nhc` object with no usable characteristics unless there is a deliberate reason. Please clarify/adjust this behavior and update the test accordingly.

3. From an idiomatic Go perspective, the main issue is exported names without doc comments: `NHCAttributeType`, the `NHCCharacteristic...` constants, `NHC`, `NHCCharacteristic`, `NHCBGPID`, `UnmarshalNHC`, and `SemanticallyMatchesNextHop`. Please add short package-style comments. The parser’s manual offset handling itself is not a blocker; that is normal for protocol decoding in Go.

Tests and CI look green, but I’d still like the mixed MP_REACH/MP_UNREACH behavior covered because that is the most likely regression point.

sbezverk (approved):

PR #475, opened 2026-09-03
add consistent and authoritative peer id managment

sbezverk:
## Summary

This change introduces consistent BMP peer identity handling based on per-peer state learned from PeerUp messages. Router identity is now tracked per BMP peer/session and attached to route/stat messages before publishing, avoiding incorrect reuse of global speaker identity.

## Changes

- Added `PeerIdentity` to carry `SpeakerIP`, `RouterIP`, `PeerIP`, hashes, and Loc-RIB state.
- Added identity extraction from PeerUp messages with safe fallback to BMP speaker IP.
- Updated producer to cache identities on PeerUp, remove them on PeerDown, and attach identity snapshots to route/stat messages.
- Updated message handlers to use per-peer identity for `router_ip` and `router_hash`.
- Added focused unit tests for normal peers, Loc-RIB, missing/malformed identities, invalid IPs, cache fallback, and PeerUp/PeerDown behavior.

## Validation

- `go test ./pkg/message`
- `go test ./...`


sbezverk:
@ZacharyPuls @Vadims06  Here is an image which is supposed to correctly provide router ip and router hash regardless of Peer types, etc. 

docker.io/sbezverk/gobmp:PR_475

Since you were contributing in this area I would appreciate if you try it and let me know if it works well in your environments. I will keep this PR not merged until I get some feedback from you. Feel free to check/review the code.

PR #476, opened 2026-09-04
Add BGP-MUP SAFI (AFI 1/2, SAFI 85) support

higebu:
Adds BGP-MUP SAFI (AFI 1/2, SAFI 85) support per draft-ietf-bess-mup-safi-01: NLRI parsing for the four 3gpp-5g route types and the ST route TLVs, the BGP MUP Extended Community (type 0x0c), a MUPPrefix message, and gobmp.parsed.mup{,_v4,_v6} topics.


sbezverk:
@higebu  Please clarify ST1 TLV handling. Section 3.1.5 marks the currently defined TLVs as applicable to ST2 and says TLVs received for a route type where they are not applicable MUST be ignored. `UnmarshalST1Route` currently decodes and stores those TLVs, and `message.mup()` publishes them for ST1. I would either drop inapplicable TLVs for ST1 or keep them only as raw observability with an explicit test/comment explaining that this is intentional for BMP export.

pkg/mup/st1.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil, fmt.Errorf("not enough data for prefix: need %d bytes, have %d", byteLen, len(b)-p)
+	}
+	r.Prefix = make([]byte, l)
+	copy(r.Prefix, b[p:p+byteLen])
+	p += byteLen
+	// TEID (4) + QFI (1) + Endpoint Address Length (1)
+	if p+6 > len(b) {
+		return nil, fmt.Errorf("not enough data for TEID, QFI and Endpoint Address Length: need 6 bytes, have %d", len(b)-p)
+	}
+	r.TEID = binary.BigEndian.Uint32(b[p : p+4])
```
sbezverk, line 82: Please reject TEID value 0 for MUP ST routes. draft-ietf-bess-mup-safi-01 says TEID value 0 is invalid/malformed for Type 1 ST, and also for Type 2 ST when a TEID is present. The current parser accepts `00 00 00 00` in ST1 and zero-valued ST2 TEIDs, so malformed routes can be published. Please add parser checks and unit tests for ST1 TEID=0, ST2 full TEID=0, and ST2 partial TEID whose significant bits are zero.

pkg/mup/st2.go
```diff
@@ -0,0 +1,112 @@
...
+	copy(r.EndpointAddress, b[p:p+l])
+	p += l
+	if teidBits := int(r.EndpointLength) - l*8; teidBits > 0 {
+		byteLen := (teidBits + 7) / 8
+		if p+byteLen > len(b) {
+			return nil, fmt.Errorf("not enough data for TEID: need %d bytes, have %d", byteLen, len(b)-p)
+		}
+		b4 := make([]byte, 4)
+		copy(b4, b[p:p+byteLen])
+		teid := binary.BigEndian.Uint32(b4)
```
sbezverk, line 75: Please reject TEID value 0 for MUP ST routes. draft-ietf-bess-mup-safi-01 says TEID value 0 is invalid/malformed for Type 1 ST, and also for Type 2 ST when a TEID is present. The current parser accepts `00 00 00 00` in ST1 and zero-valued ST2 TEIDs, so malformed routes can be published. Please add parser checks and unit tests for ST1 TEID=0, ST2 full TEID=0, and ST2 partial TEID whose significant bits are zero.

pkg/mup/mup.go
```diff
@@ -0,0 +1,190 @@
...
+		n.RouteType = binary.BigEndian.Uint16(b[p+1 : p+3])
+		n.Length = b[p+3]
+		p += 4
+		l := int(n.Length)
+		if p+l > len(b) {
+			return nil, fmt.Errorf("not enough data for MUP route type %d: need %d bytes, have %d", n.RouteType, l, len(b)-p)
+		}
+		// Only the 3gpp-5g architecture type is defined, anything else carries
+		// an unknown encoding and cannot be parsed.
+		if n.ArchitectureType != ArchType3GPP5G {
```
sbezverk, line 139: Please skip unknown/malformed MUP NLRIs where the common header length lets us advance, instead of aborting the whole MUP attribute. The draft says unknown route types should be ignored/logged and malformed NLRIs should be skipped while continuing to process the rest of the UPDATE. Returning an error here makes `processMPUpdate` drop every later valid MUP NLRI in the same MP_REACH/MP_UNREACH attribute.

pkg/mup/mup.go
```diff
@@ -0,0 +1,190 @@
...
+	if glog.V(6) {
+		glog.Infof("MUP NLRI Raw: %s, ipv6: %t, pathID: %t", tools.MessageHex(b), ipv6, pathID)
+	}
+	if len(b) == 0 {
+		return nil, ErrEmptyNLRI
+	}
+	r, err := parseMUPNLRI(b, ipv6, pathID)
+	if err == nil {
+		return r, nil
+	}
```
sbezverk: I do not think the parser should silently retry with the opposite Add-Path setting. RFC 7911 makes the extended NLRI encoding depend on the negotiated ADD-PATH capability for the corresponding `<AFI, SAFI>`. Retrying on any parse error can turn a truly malformed MUP NLRI into a plausible but shifted decode. If we need a workaround for broken devices, please make that explicit/narrow and log it, rather than accepting both encodings as normal behavior.

pkg/message/mup.go
```diff
@@ -0,0 +1,142 @@
...
+	case 0:
+		operation = "add"
+	case 1:
+		operation = "del"
+	default:
+		return nil, fmt.Errorf("unknown operation %d", op)
+	}
+
+	route, err := nlri.GetNLRIMUP()
+	if errors.Is(err, mup.ErrEmptyNLRI) {
```
sbezverk, line 27: Please only emit MUP EOR for empty MP_UNREACH_NLRI. RFC 4724 defines non-IPv4-unicast EOR as an UPDATE containing only MP_UNREACH_NLRI with no withdrawn routes for that `<AFI, SAFI>`. Right now any empty MUP NLRI error becomes a `del` EOR, even if this path was called for MP_REACH/AddPrefix. A guard on `op == DelPrefix` and/or `nlri.(*bgp.MPUnReachNLRI)` would avoid publishing invalid empty MP_REACH as a valid EOR.

pkg/mup/isd.go
```diff
@@ -0,0 +1,76 @@
...
+	r.PrefixLength = b[8]
+	l := addrLen(ipv6)
+	if int(r.PrefixLength) > l*8 {
+		return nil, fmt.Errorf("invalid prefix length %d (maximum %d)", r.PrefixLength, l*8)
+	}
+	byteLen := (int(r.PrefixLength) + 7) / 8
+	if 9+byteLen > len(b) {
+		return nil, fmt.Errorf("not enough data for prefix: need %d bytes, have %d", byteLen, len(b)-9)
+	}
+	r.Prefix = make([]byte, l)
```
sbezverk, line 49: Please canonicalize non-octet bit fields before storing/publishing them. ISD/ST1 prefixes and ST2 partial TEIDs are bit-length fields, but the code copies whole bytes, so unused padding bits can leak into the printed prefix/TEID and any future route hash. Existing L3VPN code masks unused prefix bits after copying. Please mask or reject non-zero padding and add tests for an ISD/ST1 `/25` and an ST2 endpoint length like 42.

pkg/mup/st1.go
```diff
@@ -0,0 +1,163 @@
...
+	p++
+	l := addrLen(ipv6)
+	if int(r.PrefixLength) > l*8 {
+		return nil, fmt.Errorf("invalid prefix length %d (maximum %d)", r.PrefixLength, l*8)
+	}
+	byteLen := (int(r.PrefixLength) + 7) / 8
+	if p+byteLen > len(b) {
+		return nil, fmt.Errorf("not enough data for prefix: need %d bytes, have %d", byteLen, len(b)-p)
+	}
+	r.Prefix = make([]byte, l)
```
sbezverk, line 74: Please canonicalize non-octet bit fields before storing/publishing them. ISD/ST1 prefixes and ST2 partial TEIDs are bit-length fields, but the code copies whole bytes, so unused padding bits can leak into the printed prefix/TEID and any future route hash. Existing L3VPN code masks unused prefix bits after copying. Please mask or reject non-zero padding and add tests for an ISD/ST1 `/25` and an ST2 endpoint length like 42.

pkg/mup/st2.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	r.EndpointAddress = make([]byte, l)
+	copy(r.EndpointAddress, b[p:p+l])
+	p += l
+	if teidBits := int(r.EndpointLength) - l*8; teidBits > 0 {
+		byteLen := (teidBits + 7) / 8
+		if p+byteLen > len(b) {
+			return nil, fmt.Errorf("not enough data for TEID: need %d bytes, have %d", byteLen, len(b)-p)
+		}
+		b4 := make([]byte, 4)
```
sbezverk, line 72: Please canonicalize non-octet bit fields before storing/publishing them. ISD/ST1 prefixes and ST2 partial TEIDs are bit-length fields, but the code copies whole bytes, so unused padding bits can leak into the printed prefix/TEID and any future route hash. Existing L3VPN code masks unused prefix bits after copying. Please mask or reject non-zero padding and add tests for an ISD/ST1 `/25` and an ST2 endpoint length like 42.

pkg/mup/st1.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+		r.SourceAddress = make([]byte, saLen)
+		copy(r.SourceAddress, b[p:p+saLen])
+		p += saLen
+	default:
+		return nil, fmt.Errorf("invalid source address length %d (expected 0, 32 or 128)", r.SourceAddressLength)
+	}
+	// Draft Section 3.1.5 defines no TLV applicable to Type 1 ST routes, but
+	// received TLVs are still decoded and kept so they stay observable.
+	tlvs, err := unmarshalTLVs(b[p:])
```
sbezverk: Please clarify ST1 TLV handling. Section 3.1.5 marks the currently defined TLVs as applicable to ST2 and says TLVs received for a route type where they are not applicable MUST be ignored. `UnmarshalST1Route` currently decodes and stores those TLVs, and `message.mup()` publishes them for ST1. I would either drop inapplicable TLVs for ST1 or keep them only as raw observability with an explicit test/comment explaining that this is intentional for BMP export.

pkg/message/types.go
```diff
@@ -446,6 +447,55 @@ type RTCPrefix struct {
...
+// MUPPrefix defines the structure of BGP-MUP message (AFI 1/2, SAFI 85)
+// as defined in draft-ietf-bess-mup-safi-01.
+// Which of the route type specific fields are set depends on RouteType.
+type MUPPrefix struct {
+	Key             string              `json:"_key,omitempty"`
+	ID              string              `json:"_id,omitempty"`
+	Rev             string              `json:"_rev,omitempty"`
+	Action          string              `json:"action,omitempty"` // Action can be "add" or "del"
+	Sequence        int                 `json:"sequence,omitempty"`
+	Hash            string              `json:"hash,omitempty"`
```
sbezverk, line 459: Please wire MUPPrefixinto the existing message hash population path. MUP messages are published through marshalAndPublish(), but ensureMessageHash()currently handles only peer, unicast, and L3VPN messages. As a result, the new MUP messages may be published with an emptyhash. Please add a focused test asserting that the published MUP message has a stable non-empty hash.

sbezverk (changes requested):
Thank you for your PR, please addressed review comments and provide requested clarifications. Thank you.

pkg/message/hash.go
```diff
@@ -78,6 +84,37 @@ func setL3VPNPrefixHash(m *L3VPNPrefix) {
...
+// TLVs are attributes of a route, not its identity.
+func setMUPPrefixHash(m *MUPPrefix) {
+	if m == nil || m.Hash != "" || m.IsEOR {
+		return
+	}
+	teid := ""
+	if m.TEID != nil {
+		teid = strconv.FormatUint(uint64(*m.TEID), 10)
+	}
+	m.Hash = hashParts(
```
sbezverk: setMUPPrefixHash includes EndpointAddress, EndpointLen, and TEID for every MUP route type. However, [draft Section 3.1.3](https://datatracker.ietf.org/doc/html/draft-ietf-bess-mup-safi-01#section-3.1.3) defines the ST1 route key as only RD, Prefix Length, and Prefix. If forwarding attributes such as TEID or endpoint address change while the ST1 route key remains the same, this code generates a different hash and may prevent downstream add/withdraw correlation. Please build the hash fields per route type and add a test proving that two ST1 messages with the same route key but different forwarding attributes produce the same hash.

pkg/message/mup.go
```diff
@@ -0,0 +1,146 @@
...
+			PeerHash:       ph.GetPeerHash(),
+			RemoteBGPID:    ph.GetPeerBGPIDString(),
+			PeerIP:         ph.GetPeerAddrString(),
+			PeerType:       uint8(ph.PeerType),
+			PeerASN:        ph.PeerAS,
+			Timestamp:      ph.GetPeerTimestamp(),
+			Nexthop:        nlri.GetNextHop(),
+			BaseAttributes: update.BaseAttributes,
+			IsIPv4:         !nlri.IsIPv6NLRI(),
+			IsNexthopIPv4:  !nlri.IsNextHopIPv6(),
```
sbezverk: MP_UNREACH_NLRI.IsNextHopIPv6() always returns false because withdrawals do not carry a next hop. Therefore, !nlri.IsNextHopIPv6() sets IsNexthopIPv4=true for every MUP withdrawal, including IPv6 withdrawals. Derive this field from the actual non-empty next-hop value, or explicitly handle MP_UNREACH_NLRI. Please add an IPv6 withdrawal test.

pkg/mup/isd.go
```diff
@@ -0,0 +1,77 @@
...
+		return nil, fmt.Errorf("failed to parse RD: %w", err)
+	}
+	r.RD = rd
+	r.PrefixLength = b[8]
+	l := addrLen(ipv6)
+	if int(r.PrefixLength) > l*8 {
+		return nil, fmt.Errorf("invalid prefix length %d (maximum %d)", r.PrefixLength, l*8)
+	}
+	byteLen := (int(r.PrefixLength) + 7) / 8
+	if 9+byteLen > len(b) {
```
sbezverk: The ISD parser accepts any route-specific field longer than the expected size and silently ignores the trailing bytes. Since the common-header Length defines the complete route-specific field, the parser should require 9+byteLen == len(b). Otherwise malformed ISD NLRIs can be published as valid routes. Please add a test with one trailing byte.

sbezverk (changes requested):

sbezverk:
@higebu great progress, please address a few more comments.  BGP-MUP is a new feature, are you aware of any actual bgp speaker supporting it? Or it was just a coding exercise without practical use case? Appreciate your feedback.

higebu:
@sbezverk Thanks, the remaining comments are addressed and pushed.

Not a coding exercise: GoBGP implements the SAFI (this PR's test vectors come from its serializer), and FRR is adding it in 
[FRRouting/frr#22311](https://github.com/FRRouting/frr/pull/22311)
 including bmp monitor. gobmp is the collector for that BMP feed.

sbezverk:
@higebu thank you for your feedback. I am planning to extend gobmp CI/CD integration tests and add gobgp to test bgp-mup.
It looks like copilot decided to chime in with some comments, please address them, thank you.

sbezverk:
@higebu please address 2 outstanding comments from the copilot, so I could merge it, thank you.

sbezverk:
@higebu I ran some tests with injecting MUP updates to gobgp and then using gobmp built from your PR, and found that:
The MUP route is decoded correctly, but Prefix-SID TLV type 5 is currently serialized as unknown_value. Since the repository already contains SRv6 L3 Service decoding, could this PR wire BGP Prefix-SID TLV type 5 to that decoder and add a test covering the MUP ISD/DSD attribute? This would ensure MUP support does not lose the SID and endpoint behavior semantics.

sbezverk (approved):
Approving PR, as there is no any real blockers with a tech debt issue #479

PR #477, opened 2026-09-12
Add Inter-AS BGP-LS topology support

SergeKrier:
## Summary

- Add BGP-LS Inter-AS Link NLRI type 7 decoding based on `draft-ietf-idr-bgpls-inter-as-topology-ext-44`.
- Decode and validate the Local Node Descriptor and Inter-AS Link Descriptor TLVs:
  - Remote AS Number (TLV 270)
  - IPv4 Remote ASBR ID (TLV 271)
  - IPv6 Remote ASBR ID (TLV 272)
  - Existing link identity descriptors (TLVs 258–262)
- Preserve duplicate and unknown Inter-AS link descriptors in canonical RFC 9552 wire order.
- Publish Inter-AS half-links through the existing `gobmp.parsed.ls_link` output with explicit local and remote ASBR identifiers, dual-stack link addresses, and an `is_inter_as` marker.
- Publish TLV 270 as the dedicated `remote_asn` field while reserving `remote_node_asn` for normal Link NLRIs that contain a Remote Node Descriptor.
- Preserve Add-Path IDs for BGP-LS link updates and withdrawals.
- Represent the 8-octet BGP-LS Instance Identifier as `int64`, consistently with existing LS message types.
- Publish an explicit `(local ASN, BGP-LS Instance-ID)` domain key as required by draft-44.
- Reject the OSPF Area-ID descriptor when it appears with a non-OSPF source Protocol-ID.
- Guard all new model, validation, and producer pointer dereferences against nil values.
- Parse type-7 descriptors into an ordered slice instead of a map, allowing duplicate TLVs without losing their order.
- Enforce RFC 9552 canonical NLRI ordering by type, length, and opaque value.
- Reuse existing BGP-LS link-attribute processing for TE metrics and other link properties.

## Protocol behavior

The implementation validates the mandatory draft-44 fields and fixed-length encodings, including:

- Local Autonomous System (TLV 512)
- Local IGP Router-ID (TLV 515)
- At least one local IPv4 or IPv6 ASBR Router-ID (TLV 1028 or 1029)
- Remote AS Number (TLV 270)
- At least one IPv4 or IPv6 Remote ASBR ID (TLV 271 or 272)
- Fixed lengths for link identity descriptors 258–262 when present

Both IPv4 and IPv6 interface, neighbor, and ASBR identifiers are retained when a dual-stack half-link advertises both address families.

Draft-44 adds a normative requirement for consumers to distinguish IGP domains using the `(ASN, BGP-LS Instance-ID)` tuple. The structured message therefore includes `inter_as_domain_key` with both values. Direct and Static Protocol-IDs remain accepted, and full four-octet Remote AS Numbers are decoded without applying the removed draft-38 high-order-zero restriction.

Reference: https://datatracker.ietf.org/doc/html/draft-ietf-idr-bgpls-inter-as-topology-ext-44

## Test plan

- [x] Parse a valid dual-stack Inter-AS Link NLRI.
- [x] Verify mandatory Local Node and Inter-AS Link descriptors.
- [x] Reject missing mandatory descriptors.
- [x] Reject invalid fixed-length TLVs.
- [x] Preserve duplicate unknown link descriptors and their canonical order.
- [x] Reject non-canonical descriptor ordering by type, length, or value.
- [x] Reject truncated descriptor headers and values.
- [x] Keep BGP-LS Instance Identifiers consistent with existing `int64` LS fields.
- [x] Distinguish equal Instance Identifiers in different ASes with the composite domain key.
- [x] Accept Direct and Static source Protocol-IDs from wire-format NLRIs.
- [x] Accept full four-octet Remote AS Numbers.
- [x] Reject OSPF Area-ID TLV 514 for non-OSPF sources.
- [x] Verify nil receivers, nil descriptors, nil producer inputs, and nil updates do not panic.
- [x] Publish TLV 270 as `remote_asn` without overloading `remote_node_asn`.
- [x] Publish structured Inter-AS link fields through the existing LS Link topic.
- [x] Preserve Add-Path IDs for advertisements and withdrawals.
- [x] Run `go test ./...`.
- [x] Run `go vet ./...`.
- [x] Build `gobmp`, `player`, `validator`, and `validator_v2`.


pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,208 @@
...
+}
+
+// GetIdentifier returns the unsigned 64-bit BGP-LS Instance Identifier.
+func (l *InterASLinkNLRI) GetIdentifier() uint64 {
+	return binary.BigEndian.Uint64(l.Identifier[:])
+}
+
+// GetLocalASBRIPv4 returns the local ASBR IPv4 Router-ID from TLV 1028.
+func (l *InterASLinkNLRI) GetLocalASBRIPv4() net.IP {
+	if tlv, ok := l.LocalNode.SubTLV[1028]; ok {
```
sbezverk, line 65: if LocalNode is nil, it will cause panic, please check for such condition and add unit test testing it.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,208 @@
...
+		return net.IP(tlv.Value).To4()
+	}
+	return nil
+}
+
+// GetLocalASBRIPv6 returns the local ASBR IPv6 Router-ID from TLV 1029.
+func (l *InterASLinkNLRI) GetLocalASBRIPv6() net.IP {
+	if tlv, ok := l.LocalNode.SubTLV[1029]; ok && len(tlv.Value) == net.IPv6len {
```
sbezverk, line 76: if LocalNode is nil, it will cause panic, please check for such condition and add unit test testing it.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,208 @@
...
+		return net.IP(tlv.Value).To16()
+	}
+	return nil
+}
+
+// GetRemoteASN returns the neighboring autonomous system from TLV 270.
+func (l *InterASLinkNLRI) GetRemoteASN() uint32 {
+	if tlv, ok := l.Link.LinkTLV[RemoteASNumberType]; ok && len(tlv.Value) >= 4 {
```
sbezverk, line 87: if LocalNode is nil, it will cause panic, please check for such condition and add unit test testing it.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,208 @@
...
+		return binary.BigEndian.Uint32(tlv.Value)
+	}
+	return 0
+}
+
+// GetRemoteASBRIPv4 returns the neighboring ASBR IPv4 Router-ID from TLV 271.
+func (l *InterASLinkNLRI) GetRemoteASBRIPv4() net.IP {
+	if tlv, ok := l.Link.LinkTLV[IPv4RemoteASBRIDType]; ok {
```
sbezverk, line 98: if LocalNode is nil, it will cause panic, please check for such condition and add unit test testing it.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,208 @@
...
+		return net.IP(tlv.Value).To4()
+	}
+	return nil
+}
+
+// GetRemoteASBRIPv6 returns the neighboring ASBR IPv6 Router-ID from TLV 272.
+func (l *InterASLinkNLRI) GetRemoteASBRIPv6() net.IP {
+	if tlv, ok := l.Link.LinkTLV[IPv6RemoteASBRIDType]; ok && len(tlv.Value) == net.IPv6len {
```
sbezverk, line 109: if LocalNode is nil, it will cause panic, please check for such condition and add unit test testing it.

Please make sure that 
&InterASLinkNLRI{}
&InterASLinkNLRI{LocalNode: nil, Link: nil}

do not trigger panic.

pkg/message/ls-inter-as-link.go
```diff
@@ -0,0 +1,114 @@
...
+	"fmt"
+
+	"github.com/sbezverk/gobmp/pkg/base"
+	"github.com/sbezverk/gobmp/pkg/bgp"
+	"github.com/sbezverk/gobmp/pkg/bmp"
+)
+
+// lsInterASLink maps a decoded Inter-AS half-link and its BGP-LS attributes into the shared LSLink schema.
+func (p *producer) lsInterASLink(link *base.InterASLinkNLRI, nextHop string, op int, ph *bmp.PerPeerHeader, update *bgp.Update, isIPv6 bool) (*LSLink, error) {
+	// Translate the internal operation code into the stable output action.
```
sbezverk, line 28: link is not checked for nil, please make sure all pointers are checked before dereferencing them.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,208 @@
...
+	}
+	if hasIPv6 && ipv6.Length != net.IPv6len {
+		return fmt.Errorf("Inter-AS Link Local Node Descriptor TLV 1029 has length %d, expected 16", ipv6.Length)
+	}
+	return nil
+}
+
+// validateInterASLinkDescriptors enforces remote ASBR identity and fixed-length link descriptor encodings.
+func validateInterASLinkDescriptors(link *LinkDescriptor) error {
+	// The remote ASN and at least one remote ASBR Router-ID identify the neighboring half-link.
```
sbezverk, line 211: please check pointers fo9r nil before dereferencing.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,208 @@
...
+		return fmt.Errorf("Inter-AS Link Descriptors missing IPv4 or IPv6 Remote ASBR ID TLV")
+	}
+	if hasIPv4 && ipv4.Length != net.IPv4len {
+		return fmt.Errorf("Inter-AS Link Descriptor TLV 271 has length %d, expected 4", ipv4.Length)
+	}
+	if hasIPv6 && ipv6.Length != net.IPv6len {
+		return fmt.Errorf("Inter-AS Link Descriptor TLV 272 has length %d, expected 16", ipv6.Length)
+	}
+	// Validate optional link-correlation descriptors when the source IGP advertises them.
+	lengths := map[uint16]uint16{258: 8, 259: 4, 260: 4, 261: 16, 262: 16}
```
sbezverk: allocating map in a hot path is not efficient, please use switch instead.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,208 @@
...
+	LocalNode     *NodeDescriptor
+	Link          *LinkDescriptor
+	LocalNodeHash string
+	LinkHash      string
+}
+
+// InterASDomainKey identifies an IGP domain using the tuple mandated by draft-44.
+type InterASDomainKey struct {
+	ASN        uint32 `json:"asn"`
+	Identifier uint64 `json:"identifier"`
```
sbezverk: Please make identifier consistent with other LS types. All the rest uses int64

pkg/message/ls-inter-as-link.go
```diff
@@ -0,0 +1,114 @@
...
+		Timestamp:        ph.GetPeerTimestamp(),
+		DomainID:         link.GetIdentifier(),
+		Nexthop:          nextHop,
+		PeerIP:           ph.GetPeerAddrString(),
+		Protocol:         link.GetProtocolID(),
+		ProtocolID:       link.ProtocolID,
+		LSID:             link.LocalNode.GetLSID(),
+		LocalNodeHash:    link.LocalNodeHash,
+		LocalNodeASN:     link.LocalNode.GetASN(),
+		RemoteNodeASN:    link.GetRemoteASN(),
```
sbezverk: For a normal link NLRI, this field describes the ASN associated with the remote node. For type 7, the value comes from the Inter-AS Link Descriptor and represents the remote AS, not necessarily a remote node ASN. Downstream consumers may interpret the field incorrectly.
Consider adding a distinct RemoteASN field or documenting that RemoteNodeASN means the remote AS when IsInterAS == true.

sbezverk (commented):
Please address identified issues. Please make sure that all pointers are check for nil before they are used.

pkg/message/ls-inter-as-link.go
SergeKrier: Good catch.  RemoteNodeASN  already has established semantics for a normal Link NLRI, while TLV 270 in an Inter-AS Link NLRI identifies the neighboring AS rather than a decoded remote node.
 I’ll add a dedicated  RemoteASN  field ( remote_asn ) for type 7 and leave  RemoteNodeASN  unset for Inter-AS links. This avoids overloading the existing field and prevents downstream ambiguity. I’ll update the tests and PR description accordingly.

SergeKrier: addressed

sbezverk:
@SergeKrier This PR appears intentionally limited to AFI 16388 / SAFI 71. Please add an issue for  SAFI 72 Inter-AS Link NLRIs support.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,248 @@
...
+		return nil, err
+	}
+	l.LocalNode = localNode
+	localHash := md5.Sum(localNodeBytes)
+	l.LocalNodeHash = hex.EncodeToString(localHash[:])
+	p += 4 + ndl
+	if p == len(b) {
+		return nil, fmt.Errorf("Inter-AS Link NLRI has no link descriptors")
+	}
+	// Decode all remaining Inter-AS Link Descriptors, preserving unknown TLVs in the descriptor map.
```
sbezverk, line 147: Underlying UnmarshalTLV stores descriptors in a map and rejects duplicate TLV types.

RFC 9552 requires unknown TLVs to be preserved and defines ordering rules for multiple TLVs of the same type. The current representation cannot preserve duplicate unknown descriptors or their ordering. [RFC 9552](https://www.rfc-editor.org/rfc/rfc9552.html)

To be compliant with RFC, a rfc specific tlv processing function should be used, also map should not be used to allow duplicate TLVs.

sbezverk (changes requested):
Thank you for addressing previous comments, please check and address TLVs unmarshaling logic.

SergeKrier:
> @SergeKrier This PR appears intentionally limited to AFI 16388 / SAFI 71. Please add an issue for SAFI 72 Inter-AS Link NLRIs support.

 I could not find an IETF draft that explicitly specifies Inter-AS Link NLRI type 7 for AFI 16388 / SAFI 72. Draft-ietf-idr-bgpls-inter-as-topology-ext-44 defines type 7 but does not mention BGP-LS-VPN, SAFI 72, or Route Distinguisher semantics. RFC 9552 provides the generic SAFI 72 RD-prefixed envelope, but does not establish applicability for this newer NLRI type. I will create a follow-up issue to track SAFI 72 support pending clarification or standardization. https://github.com/sbezverk/gobmp/issues/478

pkg/base/inter-as-link-nlri.go
SergeKrier: addressed

pkg/base/inter-as-link-nlri_test.go
```diff
@@ -0,0 +1,385 @@
...
+	if err := validateInterASLocalNode(nil, OSPFv2); err == nil {
+		t.Error("validateInterASLocalNode(nil) returned nil")
+	}
+	if err := validateInterASLinkDescriptors(nil); err == nil {
+		t.Error("validateInterASLinkDescriptors(nil) returned nil")
+	}
+}
+
+// TestUnmarshalInterASLinkNLRIFixedLengths verifies every fixed-width descriptor rejects invalid lengths.
+func TestUnmarshalInterASLinkNLRIFixedLengths(t *testing.T) {
```
sbezverk, line 348: TLVs 259, 260, 261, and 262 are not tested, please add malformed-length cases for those four TLVs.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,389 @@
...
+	if typ := binary.BigEndian.Uint16(b[p : p+2]); typ != LocalNodeDescriptorType {
+		return nil, fmt.Errorf("Inter-AS Link NLRI has Node Descriptor type %d, expected 256", typ)
+	}
+	ndl := int(binary.BigEndian.Uint16(b[p+2 : p+4]))
+	if p+4+ndl > len(b) {
+		return nil, fmt.Errorf("Inter-AS Link NLRI Local Node Descriptor truncated: need %d bytes, have %d", ndl, len(b)-p-4)
+	}
+	// Decode the mandatory Local Node Descriptor and retain a stable hash for correlation.
+	localNodeBytes := b[p : p+4+ndl]
+	if _, err := unmarshalCanonicalInterASTLVs(localNodeBytes[4:]); err != nil {
```
sbezverk, line 271: allocates and copies a complete ordered TLV slice only to validate ordering, then UnmarshalNodeDescriptor parses and copies the same bytes again. Please use a non-allocating ordering check or reuse the parsed representation.

Also it validates local-node TLV ordering, but then reparses the descriptor through UnmarshalNodeDescriptor, which stores nested TLVs in a map. Unknown local-node TLVs are consequently not exposed in the resulting LSLink, and duplicate local-node TLVs are rejected. RFC 9552 requires unknown NLRI TLVs to be preserved when propagating them. Please preserve the local-node opaque TLVs too, or explicitly document that this PR only preserves Inter-AS link-descriptor TLVs.

pkg/base/inter-as-link-nlri.go
```diff
@@ -0,0 +1,389 @@
...
+	if hasIPv4 && ipv4.Length != net.IPv4len {
+		return fmt.Errorf("Inter-AS Link Local Node Descriptor TLV 1028 has length %d, expected 4", ipv4.Length)
+	}
+	if hasIPv6 && ipv6.Length != net.IPv6len {
+		return fmt.Errorf("Inter-AS Link Local Node Descriptor TLV 1029 has length %d, expected 16", ipv6.Length)
+	}
+	return nil
+}
+
+// validateInterASLinkDescriptors enforces remote ASBR identity and fixed-length link descriptor encodings.
```
sbezverk, line 352: checks presence and length but accepts multiple TLVs of types 258-262 and 270-272. The accessors then use the first occurrence, while the published InterASDescriptors contains all occurrences. Two different TLV 270 or TLV 271 values would therefore be accepted but interpreted inconsistently. Please reject duplicate defined **singleton descriptors** while continuing to preserve duplicate unknown TLVs.

sbezverk (changes requested):
Please address below points, since they are important from RFC TLVs precessing semantics.

